PR #44, opened 2021-06-02
BPMN Errors + Messages

### LINKED ISSUE #38: Implement BPMN Error + Message properties
__What should we do?__

We should implement referenced error + message properties. The concept of adding new referenced elements will differ in the new implementation (cf. [design concept](https://github.com/camunda/camunda-modeler-design/issues/18)).

* New items will be added via "Create new" select option
* Select options (messages, errors) will be sorted alphanumeric

Group: `Error`
* Global Error Reference
* Name
* Code

Group: `Message`
* Global Message Reference
* Name

__Why should we do it?__

Child of #9 
Related to https://github.com/camunda/camunda-modeler-design/issues/18

pinussilvestrus:
Closes #38 

This adds basic support for
* BPMN Errors
* BPMN Messages

Some notes to the reviewer:

* The zeebe specific message properties will be added on top in another PR (cf. #50). 
* The border below the two action options of the selects won't be implemented by now
* For now we have two `ElementUtil` which a bit of code duplication. Rationale: We don't know where the ZeebePropertiesProvider will live in the future. That's why I wanted to minimize the shared code for the providers.

image

image

image

### DIFF
--- a/src/properties-panel/components/entries/Select.js
+++ b/src/properties-panel/components/entries/Select.js
@@ -1,3 +1,17 @@
+/**
+ * @typedef { { value: String, label: String } } Option
+ */
+
+/**
+ * Provides basic select input.
+ *
+ * @param {Object} props
+ * @param {String} props.id
+ * @param {String} props.label
+ * @param {Function} props.onChange
+ * @param {Array<Option>} [props.options]
+ * @param {String} props.value
+ */
 function Select(props) {
   const {
     id,
--- a/src/properties-panel/components/entries/Select.js
+++ b/src/properties-panel/components/entries/Select.js
@@ -28,13 +28,13 @@ function Select(props) {
   return (
     <div class="bio-properties-panel-select">
       <label for={ prefixId(id) } class="bio-properties-panel-label">{ label }</label>
-      <select id={ prefixId(id) } name={ id } class="bio-properties-panel-input" onInput={ handleChange }>
+      <select id={ prefixId(id) } name={ id } class="bio-properties-panel-input" onInput={ handleChange } value={ value }>
         {
-          options.map((option) => {
+          options.map((option, idx) => {
             return (
               <option
-                value={ option.value }
-                selected={ option.value === value }>
+                key={ idx }
+                value={ option.value }>
                 { option.label }
               </option>
             );
--- a/src/bpmn-properties-panel/entries/ReferenceSelect.js
+++ b/src/bpmn-properties-panel/entries/ReferenceSelect.js
@@ -1,0 +1,43 @@
+import {
+  useEffect
+} from 'preact/hooks';
+
+import {
+  query as domQuery
+} from 'min-dom';
+
+import {
+  usePrevious
+} from '../../properties-panel/hooks';
+
+import SelectEntry from '../../properties-panel/components/entries/Select';
+
+
+export default function ReferenceSelectEntry(props) {
+  const {
+    autoFocusEntry,
+    element,
+    getOptions
+  } = props;
+
+  const options = getOptions(element);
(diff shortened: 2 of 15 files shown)


andreasgeier (approved):
✅ &nbsp; LGTM

MaxTru:
I have not started reviewing the code in detail, but looking at the high-level I realized that the dropdown does not correctly work when I try to create multiple messages: the first two messages get created, but then the dropdown stays at `create new`
createNewMsg

MaxTru (changes requested):
I have not started reviewing the code in detail, but looking at the high-level I realized that the dropdown does not correctly work when I try to create multiple messages: the first two messages get created, but then the dropdown stays at `create new`
createNewMsg

pinussilvestrus:
Interesting. That's maybe because of the alphabetic ordering of the messages. I'll have a look 👍

pinussilvestrus:
Okay, it seems that the `select` component does not work properly when the order of the options got changed while focussed. That happens when adding a new message/error => order got alphabetized with the new option - added in between.

To keep it simple: I'd simply not sort alphabetically as long we don't have a proper custom select component. Wdyt @andreasgeier @MaxTru?

andreasgeier:
When we introduced lists we defined the general rule to set the focus to the new item after creation. We should do the same here, particularly if that helps us to keep the sort order (I don't want to lose it).

I would accept that we have this behavior also for delete (<none> selection) and selection of existing messages/errors.

pinussilvestrus:
I can imagine that this list behavior (auto focus) will be hard to implement for select entries, since the select + input entry technically don't know each other so far. I'll have to double check it.

Until then, the PR is blocked.

pinussilvestrus:
With https://github.com/bpmn-io/properties-panel-next/pull/44/commits/c68f59ff166c9aba5f8f95abfc3c0d702e41c14a I added the autofocus behavior to the reference selects when adding new elements. As discussed, I also updated the label to "Create new ..." to better indicate an action.

Kapture 2021-06-04 at 14 45 53

Since it's hard to test the focus, I didn't add a test case for this (like for list items).

That's ready for review (again) @andreasgeier @MaxTru

src/properties-panel/components/entries/Select.js
```diff
@@ -1,3 +1,16 @@
...
+ * @typedef { { value: String, label: String } } Option
+ */
+
+/**
+ * Provides basic select input.
+ *
+ * @param {Object} props
+ * @param {String} props.id
+ * @param {String} props.label
+ * @param {Array<Option>} [props.options]
```
MaxTru, line 11: `onChange` is missing

MaxTru:
The likelihood decreased, but the select sometimes still does not create a new message or error for me:

createNewMsg2

pinussilvestrus:
> The likelihood decreased, but the select sometimes still does not create a new message or error for me:
> 
> createNewMsg2

What a nasty bug, I wasn't able to reproduce by adding 30+ messages. Maybe the `preact` select component itself comes with some hickups. 

I will maybe add a stress test and see whether there is some way to reproduce it in a reliable way. I wouldn't block this PR if we see this bug recurring, but handle it in a separate issue.

pinussilvestrus:
@MaxTru  By the way now while seeing your gif, the " Create new" option still has the old label. In the current state, it is "Create new ...". Also the autofocus behavior is not visible.

Maybe you didn't update the branch?

src/properties-panel/components/entries/Select.js
pinussilvestrus: done.

MaxTru (approved):
:clap: nice

PR #53, opened 2021-06-03
chore: add codecov support

### LINKED ISSUE #52: Adding codecov
__What should we do?__

Since the codebase and the properties support becomes bigger, we should add code coverage (as codecov) to this library.

__Why should we do it?__

Like we have in other projects, it would increase our confidence we also cover different edge cases.

Cf. how we do it in form-js: https://github.com/bpmn-io/form-js/commit/999afdc81983ff868011cae2d2625603427c74e0

pinussilvestrus:
Closes #52

Note: I added `lcov` instead of `lcovonly` as a reporter, because I like the HTML tables in the report :-) 

image


nikku (approved):
:+1:

PR #57, opened 2021-06-04
Implement Message Subscription Correlation Key

### LINKED ISSUE #50: Implement Message Subscription Correlation Key
__What should we do?__

We should add the zeebe specific subscription correlation key property.

image

We have a special case here where we have to add another property inside the ZeebePropertiesProvider to the existing BpmnPropertiesProvider.

__Why should we do it?__

Requires https://github.com/bpmn-io/properties-panel-next/pull/44
Child of #9

pinussilvestrus:
Closes #50 
Builds on top of #44 

image


andreasgeier (approved):
✅ &nbsp; LGTM

MaxTru (approved):
:wave:

PR #60, opened 2021-06-07
fix(bpmn-properties-panel): do not select deleted elements

### LINKED ISSUE #20: Selected Element don't get updated on deletion
__Describe the Bug__

When the current element is removed from the canvas, the properties panel don't receive an update. The deleted element is kept.

Kapture 2021-05-19 at 14 23 42

__Steps to Reproduce__

1. npm start
2. select the task
3. delete the task => task is still in the context

__Expected Behavior__

The task is not selected anymore. Preferably, the root element is selected.

__Environment__

- Host (Browser/Node version), if applicable: Chrome 90.0.4430.212
- OS: macOS 10.15.
- Library version: current `main`

pinussilvestrus:
This prevents that we react to element updates for deleted elements. This is necessary because there might `element.changed` updates when an element got removed.

Closes #20

### DIFF
--- a/src/bpmn-properties-panel/BpmnPropertiesPanel.js
+++ b/src/bpmn-properties-panel/BpmnPropertiesPanel.js
@@ -32,9 +32,9 @@ export default function BpmnPropertiesPanel(props) {
     layoutConfig
   } = props;
 
-  const eventBus = injector.get('eventBus');
-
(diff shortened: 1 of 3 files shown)


philippfromme (approved):
👏🏻

PR #61, opened 2021-06-08
Some Housekeeping + Cleanups

pinussilvestrus:
This PR
* ensures we always use the standard `.bio-properties-panel-input` in all entries
* bumps dependencies
* cleanups type defs in the core part


pinussilvestrus:
We could potentially also add type generation + linting as we have in form-js. I will add an issue for that.

MaxTru (approved):
:wave:

PR #70, opened 2021-06-24
Data Markers

### LINKED ISSUE #23: Add content marker for (normal) groups

With the new properties panel, it should be easier to discover which groups are containing data and which not.

image

image

As a general rule of thumb, when creating a new element, all groups shall be _not_ marked as dirty (except for the general group, because the id is always filled).

We would like to indicate the dirty state via a visual representation (like we have it for the badge count). The "bold" / "greyed out" state should be gone.

We consider following states as "dirty" (tbd):
* TextField   => value !== empty
* TextArea => value !== empty
* Checkbox => checked != false
* Toggle Switch => checked != false
* Select => value !== empty
* Propagate all child => checked != false

Mark groups as "empty" when there is nothing configured.

/

__Additional context__

With https://github.com/bpmn-io/properties-panel-next/pull/19 we already introduce data markers for lists (count badges).
Zeplin: https://app.zeplin.io/project/5e149dc7ef8eaa51b52af327/dashboard?seid=60d2feb861f9d712892eb17e

pinussilvestrus:
Closes #23 

* moves list count markers to the right
* introduces data markers for normal groups - general approach: a group is `edited` when at least one entry is `edited`
* improves title attributes with data markers for header titles as a quick fix ==> Note: we will handle the whole `title` story on the general design polishing!

image

### DIFF
--- a/src/bpmn-properties-panel/provider/bpmn/properties/ErrorProps.js
+++ b/src/bpmn-properties-panel/provider/bpmn/properties/ErrorProps.js
@@ -6,8 +6,9 @@ import {
   sortBy
 } from 'min-dash';
 
-import TextField from '../../../../properties-panel/components/entries/TextField';
+import TextField, { isEdited as textFieldIsEdited } from '../../../../properties-panel/components/entries/TextField';
 import ReferenceSelect from '../../../entries/ReferenceSelect';
+import { isEdited as selectIsEdited } from '../../../../properties-panel/components/entries/Select';
 
 import {
   useService
@@ -37,14 +38,26 @@ export function ErrorProps(props) {
   const error = getError(element);
 
   let entries = [
-    { id: 'errorRef', component: <ErrorRef element={ element } /> }
+    {
+      id: 'errorRef',
+      component: <ErrorRef element={ element } />,
+      isEdited: selectIsEdited
+    }
   ];
 
   if (error) {
     entries = [
       ...entries,
-      { id: 'errorName', component: <ErrorName element={ element } /> },
-      { id: 'errorCode', component: <ErrorCode element={ element } /> }
+      {
+        id: 'errorName',
+        component: <ErrorName element={ element } />,
+        isEdited: textFieldIsEdited
+      },
+      {
+        id: 'errorCode',
+        component: <ErrorCode element={ element } />,
+        isEdited: textFieldIsEdited
+      }
     ];
   }
 
--- a/src/properties-panel/PropertiesPanel.js
+++ b/src/properties-panel/PropertiesPanel.js
@@ -28,7 +28,8 @@ function renderGroup(props) {
 /**
  * @typedef { {
  *    component: import('preact').ComponentChild,
- *    id: String
+ *    id: String,
+ *    isEdited?: Function
  * } } EntryDefinition
  *
  * @typedef { {
--- a/src/properties-panel/components/Group.js
+++ b/src/properties-panel/components/Group.js
@@ -1,8 +1,17 @@
 import {
+  useEffect,
   useState
 } from 'preact/hooks';
 
 import classnames from 'classnames';
+
+import {
+  query as domQuery
+} from 'min-dom';
+
+import {
+  isFunction
+} from 'min-dash';
 
 import { GroupArrowIcon } from './icons';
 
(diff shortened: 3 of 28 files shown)


pinussilvestrus:
> * improves title attributes with data markers for header titles as a quick fix ==> Note: we will handle the whole `title` story on the general design polishing!

We got some feedback, as an example, to give the data markers themselves a `title` attribute and increase the interaction box. We decided to not deal with this at the moment, but handle the general `title` story on a general design polishing task + when we decide on our general a11y strategy. 

Giving _one_ component a `title` would also mean we should handle all other icons (as add, remove, arrows, ...). We should do this in a single, dedicated step. That's why we go with updating the existing `title` for the group header title.

src/properties-panel/PropertiesPanel.js
```diff
@@ -28,7 +28,8 @@ function renderGroup(props) {
 /**
  * @typedef { {
  *    component: import('preact').ComponentChild,
- *    id: String
+ *    id: String,
+ *    isDirty?: Function
```
MaxTru, line 32: in the PR we use the term `dirty` to indicate that content exists.

However, from my understanding `dirty` would typically refer to a state where content exists, which is not persisted or synchronzied (for example see our dirty diagram state).

As alternative we could go for `containsContent`, or just turn it around and go for `empty`. Wdyt?

src/properties-panel/components/Group.js
```diff
@@ -40,4 +79,17 @@ export default function Group(props) {
...
+}
+
+function DataMarker() {
+  return (
+    <div class="bio-properties-panel-dot">
+      <svg
+        aria-label="dirty" role="img" xmlns="http://www.w3.org/2000/svg"
+        viewBox="0 0 100 100"
+      >
+        <circle fill="currentColor" cx="50" cy="50" r="50" />
```
MaxTru, line 91: Would we not rather style the color of the circle explicitly via css? This is where I would look for the style of it.

src/bpmn-properties-panel/provider/bpmn/properties/ErrorProps.js
```diff
@@ -37,14 +38,26 @@ export function ErrorProps(props) {
   const error = getError(element);
 
   let entries = [
-    { id: 'errorRef', component: <ErrorRef element={ element } /> }
+    {
+      id: 'errorRef',
+      component: <ErrorRef element={ element } />,
+      isDirty: selectIsDirty
```
MaxTru, line 44: I like the architecture of the solution and how flexible we can add the isDirty function

src/properties-panel/components/Group.js
```diff
@@ -20,12 +29,42 @@ export default function Group(props) {
 
   const toggleOpen = () => setOpen(!open);
 
+  const [ dirty, setDirty ] = useState(false);
+
+  // set dirty state depending on all entries
+  useEffect(() => {
+    const filledEntries = entries.filter(entry =>{
```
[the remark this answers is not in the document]
MaxTru: (minor detail) We could also use `.findIndex` or similar (we don't need to iterate all entries, just search if a non-empty on exists and then we can stop)

MaxTru (commented):
Nice, I like the architecture / how flexible we can add the `isDirty` check

Just some smaller comments.

src/properties-panel/components/Group.js
pinussilvestrus: I simply re-use the dot from the Camunda Modeler, to avoid different styles of doing it.

https://github.com/camunda/camunda-modeler/blob/develop/client/src/plugins/version-info/VersionInfo.js#L132

src/properties-panel/PropertiesPanel.js
pinussilvestrus: > persisted or synchronzied

That's true we use this term differently in other places. Maybe a better term would be `isEdited`, like we have for the title attribute.

src/properties-panel/components/Group.js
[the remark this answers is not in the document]
pinussilvestrus: Could be a nice performance improvement 👍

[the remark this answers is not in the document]
pinussilvestrus: Changed to `find` + stops when the first edited entry was found.

src/properties-panel/PropertiesPanel.js
pinussilvestrus: I changed the term `dirty` to `edited`.

src/properties-panel/components/Group.js
pinussilvestrus: Ah, now I got what you meant. I simply used the default font color here (`currentColor`). We could decide to state `black` in the selector, but since we don't have a fixed color palette yet, I'd keep it to the default and change it once we do the design polishing.

andreasgeier: Just a comment: the unread marker for versions and the data marker for the properties panel are not the same UI component. Although they look similar (equal shape but different colors) this may change as we learn more about how they are perceived.

MaxTru (approved):
:wave:

andreasgeier (approved):
✅ &nbsp; LGTM

PR #74, opened 2021-06-29
chore(properties-panel): assign key per group component

pinussilvestrus:
Not fixing any bug, but it's generally good advice to assign unique keys to child components.


pinussilvestrus:
It seems to actually fix https://github.com/bpmn-io/properties-panel-next/issues/63 on the fly. Not sure how to create a regression test for this unless creating a proper integration test...

barmac (approved):
LGTM

PR #75, opened 2021-07-01
ID Validation

### LINKED ISSUE #27: Add minimal validation to id property
__What should we do?__

As a follow to adding the ID property to the BPMN provider (cf. https://github.com/bpmn-io/properties-panel-next/pull/21) we should add minimal validation to ensure we can't blow up diagrams with this property. That includes as a minimum that missing IDs can't be committed, the last valid input value should be restored. 

We could also consider bringing back the [QName validation](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/lib/Utils.js#L168) - or decide that this is part of linting.

**tbd:** How shall the validation workflow look like?

__Why should we do it?__

Cf. https://github.com/camunda/camunda-modeler-design/issues/14

pinussilvestrus:
Closes #27 

* adds validation for ids
* adds basic validation styling (cf. [styles from Zeplin](https://app.zeplin.io/project/5e149dc7ef8eaa51b52af327/screen/60d9619fbeb59812106abf00))

Note: I was thinking about adding validation as a general concern for each entry type, but since we currently only validate this particular property and will add [linting as the general validation feature](https://github.com/camunda/camunda-modeler-design/issues/14), I decided to keep it simple and only support it for text fields for now. We can easily add validation for more entry types later.

image

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -2,9 +2,13 @@
  * Theming
  */
 .bio-properties-panel {
-  --silver-darken-80: #cdcdcd;
+  --color-silver-darken-80: #cdcdcd;
 
-  --blue-base-65: #4d90ff;
+  --color-blue-205-100-95: #e5f4ff;
+  --color-blue-205-100-50: #0095ff;
+
+  --color-red-360-100-97: #fff0f0;
+  --color-red-360-100-45: #e60000;
 
   --color-cccccc: #cccccc;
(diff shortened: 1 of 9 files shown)


pinussilvestrus:
@barmac I updated the styles to use semantic names (at least for the affected parts). We will need to do more stuff when the color palette is defined (cf. https://github.com/bpmn-io/properties-panel-next/issues/15)

barmac (approved):

PR #76, opened 2021-07-01
feat(entries): add number field component

pinussilvestrus:
Not used by any property so far, but could be important to have it in the future.

### DIFF
--- a/src/components/entries/NumberField.js
+++ b/src/components/entries/NumberField.js
@@ -12,6 +12,7 @@ function NumberField(props) {
     max,
     min,
     onInput,
+    step,
     value = ''
   } = props;
 
@@ -43,6 +44,7 @@ function NumberField(props) {
         max={ max }
         min={ min }
         onInput={ handleInput }
+        step={ step }
         value={ value } />
     </div>
   );
@@ -60,6 +62,7 @@ function NumberField(props) {
  * @param {String} props.max
  * @param {String} props.min
  * @param {Function} props.setValue
+ * @param {String} props.step
  */
 export default function NumberFieldEntry(props) {
   const {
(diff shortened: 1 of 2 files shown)


pinussilvestrus:
Let's wait with merging this when the split up is done (cf. #5)

pinussilvestrus:
Ready to review again.

barmac:
Perhaps we should also add `step` if we want to make it complete. Apart from that it looks good.

pinussilvestrus:
> Perhaps we should also add `step` if we want to make it complete. Apart from that it looks good.

Added 👍 

I updated the reviewers due to the holidays.

src/components/entries/NumberField.js
```diff
@@ -0,0 +1,109 @@
...
+  const handleInput = useMemo(() => {
+    return debounce(event => {
+
+      const {
+        validity,
+        value
+      } = event.target;
+
+      if (validity.valid) {
+        onInput(value ? parseInt(value, 10) : undefined);
```
MaxTru, line 28: By for example setting `step` to something like `0.01` the user would also be able to insert floats into the number field (e.g., in the example `10.12` would be a valid input). Still the parseInt will cut off the decimal places (so in the example `10` would be passed to `onInput`). Any reason we do this? Alternatively we could use `parseFloat` right?

MaxTru (changes requested):

src/components/entries/NumberField.js
pinussilvestrus: Good point, I didn't think about it.

If we use `parseFloat`, would that lead to always commit floating point numbers? Let's try it out.

image

pinussilvestrus: Seems like we're good to go with this, also ensuring we can have floating-point numbers via `step`.

pinussilvestrus: Fixed via https://github.com/bpmn-io/properties-panel/pull/76/commits/b81eedbfc88be75b7c9d877ac217bc9c36a7d784

MaxTru (approved):

Issue #45, opened 2021-06-02
Use <button> for buttons in the group entry


barmac:
These two are semantically buttons:

![image](https://user-images.githubusercontent.com/28307541/120455312-f355e180-c394-11eb-8d6a-6fba6f1b905d.png)

![image](https://user-images.githubusercontent.com/28307541/120455369-fd77e000-c394-11eb-9ce4-43aa3feb92a3.png)

When we use `\<div>`, it's not accessible via keyboard.

_Originally posted by @barmac in https://github.com/bpmn-io/properties-panel-next/issues/37#issuecomment-852862920_

PR #91, opened 2021-08-19
feat: use semantic HTML for buttons

barmac:
Closes https://github.com/bpmn-io/properties-panel/issues/45

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -173,14 +173,6 @@
   margin-right: 0;
 }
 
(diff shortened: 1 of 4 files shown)


marstamm (approved):
+1 for buttonized buttons

andreasgeier (approved):
I agree to use buttons to enable keyboard access and tab order as long as this does not block us in the UI – in that case, ARIA is a valid alternative.

nikku:
Very nice improvement @barmac :medal_military:

Issue #92, opened 2021-08-10
Implement list entry


barmac:
Cf. https://github.com/bpmn-io/bpmn-properties-panel/issues/2#issuecomment-895297079

barmac:
I've decided to implement it in two PRs so that we can move on with the migration. Ordering will be implemented separately.

PR #93, opened 2021-08-25
Implement list entry

barmac:
Related to #92

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -10,6 +10,8 @@
   --color-red-360-100-97: #fff0f0;
   --color-red-360-100-45: #e60000;
 
+  --color-grey-225-10-35: hsl(225, 10%, 35%);
+
   --color-cccccc: #cccccc;
   --color-aaaaaa: #aaaaaa;
   --color-white: white;
@@ -215,11 +217,6 @@
 .bio-properties-panel-list.open {
   display: block;
   margin-bottom: 6px;
-}
-
-.bio-properties-panel-list-item .bio-properties-panel-collapsible-entry-entries {
-  padding-left: 15px;
-  margin-bottom: 10px;
 }
 
 .bio-properties-panel-list-badge {
@@ -453,6 +450,8 @@ textarea.bio-properties-panel-input {
 }
 
 .bio-properties-panel-collapsible-entry-entries  {
+  padding-left: 15px;
+  margin-bottom: 10px;
   position: relative;
   display: none;
 }
@@ -497,6 +496,7 @@ textarea.bio-properties-panel-input {
   visibility: hidden;
 }
 
+.bio-properties-panel-list-entry-item:hover .bio-properties-panel-remove-entry,
 .bio-properties-panel-collapsible-entry-header:hover .bio-properties-panel-remove-entry {
   visibility: visible;
 }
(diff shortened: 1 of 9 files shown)


barmac:
This does not include ordering (which will be added in a separate PR), and can be tested with `bpmn-io/bpmn-properties-panel#48-migrate-listener-group`.

src/components/entries/List.js
```diff
@@ -0,0 +1,156 @@
...
+          const key = getKey(item);
+
+          return (<li class="bio-properties-panel-list-entry-item" key={ key }>
+            {renderItem(item, index)}
+            {
+              onRemove && (
+                <button
+                  type="button"
+                  class="bio-properties-panel-remove-entry"
+                  onClick={ () => onRemove && onRemove(item) }
```
pinussilvestrus, line 140: I wonder whether we will [be able to use `context`](https://github.com/bpmn-io/bpmn-properties-panel/blob/main/src/provider/camunda-platform/properties/FieldInjectionProps.js#L75) when we only allow callbacks here.

barmac: Tested in battle: https://github.com/bpmn-io/bpmn-properties-panel/blob/48-migrate-listener-group/src/provider/camunda-platform/properties/ListenerProps.js#L81

pinussilvestrus: 🚀 👏

src/hooks/useKeyFactory.js
```diff
@@ -0,0 +1,39 @@
...
+ *     })
+ *   }</ol>);
+ * }
+ * ```
+ *
+ * @param {any[]} dependencies
+ * @returns {(element: object) => string}
+ */
+export function useKeyFactory(dependencies = []) {
+  const map = useMemo(() => new WeakMap(), dependencies);
```
pinussilvestrus, line 25: Seems like [we're good](https://caniuse.com/?search=WeakMap) to use `WeakMap` here, only older versions of our supported browsers have problems here.

src/hooks/useKeyFactory.js
```diff
@@ -0,0 +1,39 @@
...
+ *       return <li key={key}>obj.name</li>
+ *     })
+ *   }</ol>);
+ * }
+ * ```
+ *
+ * @param {any[]} dependencies
+ * @returns {(element: object) => string}
+ */
+export function useKeyFactory(dependencies = []) {
```
pinussilvestrus, line 24: Not sure but just want to double-check: is there a need for a cleanup handler for this custom hook? Probably not because list entries are created from scratch once the selected element changes.

test/spec/components/List.spec.js
```diff
@@ -0,0 +1,778 @@
...
+        label: 'Item 3'
+      }
+    ];
+
+    const {
+      container,
+      rerender
+    } = createListEntry({ container: parentContainer, items });
+
+    items.splice(0, 1);
```
pinussilvestrus, line 671: Maybe we can test the usage of a `onRemove` handler here? A `rerender` would probably be needed as well because the items change, but at least we can verify the `onRemove` handler works as expected.

pinussilvestrus (commented):
Only minor things after first review. Otherwise, I like the way the list entry is implemented 👍  

Looks super clean so far and the concept of the custom `useKeyFactory` hook to make sure keys are always unique for non-id elements is pretty interesting.

src/hooks/useKeyFactory.js
pinussilvestrus: I think weak maps are okay here because we don't expect these maps to have a long life span, right? Did you consider any other types of maps to store the information?

andreasgeier:
The dot that indicates the start of the second level list is at the wrong place. It should be next to the list headline.

Expected vs implemented:

Screen Shot 2021-09-03 at 11 47 38

Screen Shot 2021-09-03 at 11 48 05

src/hooks/useKeyFactory.js
barmac: No, because when the callback (`getKey`) is garbage collected, so is the map. It will be recreated from scratch once we render a new component.

andreasgeier:
The create event is not consistent with the list of the first level.

Expected behavior:

- on create, insert new item on top _in expanded state and focus on the first input_
- on create, _do not affect other list items_. They keep their state, whether if they are opened or closed


https://user-images.githubusercontent.com/938244/131988403-d5961c3f-c3ae-4134-8ecd-86bb39da1567.mov

src/hooks/useKeyFactory.js
barmac: I think I will switch to a `Map` so that we can also handle primitive values.

test/spec/components/List.spec.js
barmac: OK I will do that.

barmac: Done.

src/hooks/useKeyFactory.js
barmac: Done.

barmac:
I've fixed the dot position. Regarding the behaviors, let's discuss it on Monday.

assets/properties-panel.css
```diff
@@ -10,6 +10,8 @@
   --color-red-360-100-97: #fff0f0;
```
pinussilvestrus, line 10: The indentation line (left) for the nested entries is not long enough. It should be fit to the last entry's input.

image

barmac: image

Implemented via https://github.com/bpmn-io/properties-panel/pull/93/commits/eca57f97052a401de8a702856d36d0d1409da1b8

barmac:
@andreasgeier @pinussilvestrus Ready for review. Includes sorting.

src/components/entries/List.js
```diff
@@ -0,0 +1,208 @@
...
+ * @param {object} props
+ * @param {string} props.id
+ * @param {*} props.element
+ * @param {Function} props.onAdd
+ * @param {Function} props.renderItem
+ * @param {string} [props.label='<empty>']
+ * @param {Function} [props.onRemove]
+ * @param {Item[]} [props.items]
+ * @param {boolean} [props.open]
+ * @param {(a: Item, b: Item) => -1 | 0 | 1} [props.compareFn]
```
pinussilvestrus, line 33: Can we have any reasonable default here? Probably not, right? Because

* every item has a different structure (name, label, value, ...)
* the absence of the `compareFn` indicates disabling of the sorting

pinussilvestrus (commented):
The sorting looks good now. The only thing that is missing is the autoFocus behavior when adding a new item. Will that be added on top?

Cf. @andreasgeier   

> * on create, insert new item on top _in expanded state and **focus on the first input**_

From the first level lists: 

Kapture 2021-09-10 at 08 54 45

barmac:
> The sorting looks good now. The only thing that is missing is the autoFocus behavior when adding a new item. Will that be added on top?

I missed this one. I will need to add it together with a test case.

src/components/entries/List.js
barmac: I don't think it's possible. Opposed to ListGroup, we don't know the structure of `Item` upfront. Thus, it's impossible to predict a default property for sorting.

pinussilvestrus: That's fine 👍

pinussilvestrus (approved):
Works and looks good for me now 👍  Good job 👏

andreasgeier (approved):
I was not able to run it bug-free on my machine. But @barmac and I did a review via Zoom together. So I can approve it 👍

pinussilvestrus:
🎉

PR #94, opened 2021-09-07
Simple text component

pinussilvestrus:
Adds a simple text component (without labels, descriptions, ...)

Required by https://github.com/bpmn-io/bpmn-properties-panel/pull/100 (Inputs & Outputs Lists)

image

### DIFF
--- a/src/components/entries/Simple.js
+++ b/src/components/entries/Simple.js
@@ -1,0 +1,61 @@
+import {
+  useMemo
+} from 'preact/hooks';
+
+/**
+ * @param {Object} props
+ * @param {Function} props.debounce
+ * @param {Boolean} [props.disabled]
+ * @param {Object} props.element
+ * @param {Function} props.getValue
+ * @param {String} props.id
+ * @param {Function} [props.onBlur]
+ * @param {Function} [props.onFocus]
+ * @param {Function} props.setValue
+ */
+export default function Simple(props) {
+  const {
+    debounce,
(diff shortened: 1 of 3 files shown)


pinussilvestrus:
I decoupled the branch from the nested lists so that it can be reviewed independently. Ready for review though.

src/components/entries/Simple.js
```diff
@@ -0,0 +1,57 @@
+import {
+  useMemo
+} from 'preact/hooks';
+
+/**
+ * @param {Object} props
+ * @param {Boolean} props.debounce
```
MaxTru, line 7: Isn't debounce a Function?

src/components/entries/Simple.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+export default function Simple(props) {
+  const {
+    element,
+    id,
+    debounce,
+    disabled,
+    getValue,
+    setValue
+  } = props;
```
MaxTru, line 22: Why don't we get `onFocus` and `onBlur` explicitly here?

pinussilvestrus: Good catch, there is no reason for that. I will update 👍

src/components/entries/Simple.js
pinussilvestrus: woops

MaxTru (approved):

Issue #90, opened 2021-08-06
Persist open state for list items during session


pinussilvestrus:
__Is your feature request related to a problem? Please describe__

_This is a follow up of: https://github.com/bpmn-io/properties-panel/issues/86_

Introducing #86, list items are completely unique per id. Therefore they have their own state depending on what element is selected. Since the items are unique per id, they got rerendered when a new element got selected.+

That causes that the state got lost when switching from one element to another and switch back again (the item is closed, no matter it was opened after the switch).

![Kapture 2021-08-06 at 10 55 19](https://user-images.githubusercontent.com/9433996/128484998-68cc37e5-fb10-4086-bea5-543faf7679c6.gif)

__Describe the solution you'd like__

For a smoother editing experience and to avoid opening list items multiple times during element switches, it should be possible to persist the opened state per item during a session (or multiple sessions). 

We have the [`LayoutContext`](https://github.com/bpmn-io/properties-panel/blob/main/src/context/LayoutContext.js) which was meant for exactly this purpose, but we are currently not using it.

__Describe alternatives you've considered__

* keep things as they are
* persist layout not only during runtime, but also make it possible to persist them anywhere, cf. [`propertiesPanel.layoutChanged`](https://github.com/bpmn-io/bpmn-properties-panel/blob/main/src/render/BpmnPropertiesPanel.js#L143).

__Additional context__

Cf. #86

pinussilvestrus:
A sketch of how to use the `LayoutContext` to persist the open state (cf. [this branch](https://github.com/bpmn-io/properties-panel/tree/90-persisting-item-state))

```js
import {
  useContext,
  useMemo,
  useState
} from 'preact/hooks';

import {
  LayoutContext
} from '../../context';

export default function CollapsibleEntry(props) {

  const {
    // ...
    open: shouldOpen
  } = props;

  const {
    layout,
    setLayoutForKey
  } = useContext(LayoutContext);

  const layoutKey = useMemo(() => {
    return `collapsible-${id}`;
  }, [ id ]);

  const collapsibleLayout = layout[layoutKey] || {};

  const [ open, setOpen ] = useState(shouldOpen || collapsibleLayout.open);

  const toggleOpen = () => {
    setOpen(!open);
    setLayoutForKey(layoutKey, {
      ...collapsibleLayout,
      open: !open
    });
  };

  // ...
}
```

Open for discussion: how should the layout API look like? how should the props be stored?

[...]

PR #99, opened 2021-09-24
feat(hooks): add `useLayoutState`

### LINKED ISSUE #98: Open state of groups is not persisted between element changes
__Describe the Bug__

When opening a group that is not present after an element switch, its `open` state got lost.

Kapture 2021-09-23 at 12 59 04

__Steps to Reproduce__

1. Select the root process element
2. Open "History cleanup" & "External task" groups
3. Switch to another element, note that groups are not present
4. Switch back to process =>  "History cleanup" & "External task" groups are not open anymore

__Expected Behavior__

Group `open` state got persisted across selected element changes, although they are not present in the meantime.

We could potentially use the [`LayoutContext API`](https://github.com/bpmn-io/properties-panel/blob/main/src/context/LayoutContext.js) for that.

pinussilvestrus:
Proposal: This adds a `useLayoutState` hook to persist state inside the global `LayoutContext`. The `LayoutContext` can be used to persist state for components that might not be rendered sometimes (example groups), but we want to keep.

```js
const groupId = 'general';

const [ open, setOpen ] = useLayoutState(
  [ 'groups', id, 'open' ],
  false
);

setOpen(true);
```

Potentially it would also be possible to persist this layout state outside of the application (e.g. the Camunda Modeler), via [`propertiesPanel.layoutChanged`](https://github.com/bpmn-io/bpmn-properties-panel/blob/main/src/render/BpmnPropertiesPanel.js#L142) + the [`layoutConfig`](https://github.com/bpmn-io/bpmn-properties-panel/blob/main/src/render/BpmnPropertiesPanel.js#L25) parameter.

Closes #98

Kapture 2021-09-24 at 10 46 30

### DIFF
--- a/src/PropertiesPanel.js
+++ b/src/PropertiesPanel.js
@@ -3,6 +3,11 @@ import {
   useEffect
 } from 'preact/hooks';
 
+import {
+  get,
+  set
+} from 'min-dash';
+
 import classnames from 'classnames';
 
 import Header from './components/Header';
@@ -82,16 +87,18 @@ export default function PropertiesPanel(props) {
     }
   }, [ layout, layoutChanged ]);
 
+  const getLayoutForKey = (key, defaultValue) => {
+    return get(layout, key, defaultValue);
+  };
+
   const setLayoutForKey = (key, config) => {
-    setLayout({
-      ...layout,
-      [key]: config
-    });
+    setLayout(set(layout, key, config));
   };
 
   const layoutContext = {
     layout,
     setLayout,
+    getLayoutForKey,
     setLayoutForKey
   };
 
--- a/src/components/Group.js
+++ b/src/components/Group.js
@@ -13,6 +13,10 @@ import {
   isFunction
 } from 'min-dash';
 
+import {
+  useLayoutState
+} from '../hooks';
+
 import { GroupArrowIcon } from './icons';
 
 /**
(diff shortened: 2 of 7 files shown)


nikku:
To understand the overall direction we're heading UI-wise first:

Layout state (collapsed/expanded) is sticky across elements: Section `foo` is open on element `A`, it will also be open on element `B`, if it exists there.

This PR adds the foundations to ensure that if element `B` does not contain `foo` the section is still opened in `A` once the user switches back?

pinussilvestrus:
> To understand the overall direction we're heading UI-wise first: Layout state (collapsed/expanded) is sticky across elements: Section `foo` is open on element `A`, it will also be open on element `B`, if it exists there.
> 
> This PR adds the foundations to ensure that if element `B` does not contain `foo` the section is still opened in `A` once the user switches back?

Exactly!

pinussilvestrus:
Adding @andreasgeier for a design review (proper behavior). Command to try it out

```sh
npx @bpmn-io/sr bpmn-io/bpmn-properties-panel#main -l bpmn-io/properties-panel#98-keep-group-state -c 'npm run start:platform'
```

andreasgeier:
The root level (sections/groups) works exactly as expected - great work!

But we have inconsistency on other levels. For instance, if a user creates 3 input variables and some of them are in opened state. If the input group is closed and re-opened without switching the element, the variables keep their state but when switching the element selection, all variables within the input section are closed.

@pinussilvestrus Let me know how to proceed here. Having this behavior available on the root is already a great improvement. We can discuss if you want to handle other levels as well (and how) or tackling it in another PR or ignore it for now.

pinussilvestrus:
I think we have a good chance to use the same strategy for other components as well (as sketched for list items in https://github.com/bpmn-io/properties-panel/issues/90#issuecomment-894227348). 

I think this PR is a good starting point how to tackle a similar set of problems (persisting component state during a session). If we agree on something (+ merge) we can freely do the same for other parts (as mentioned in https://github.com/bpmn-io/properties-panel/pull/99#issuecomment-932011709).

Issue #96, opened 2021-09-15
Allow `auto-open` when `shouldSort=false` for ListGroup


MaxTru:
__Is your feature request related to a problem? Please describe__

Currently a list group will not handle effects, hence also not autoOpen, when `shouldSort=false`. See [here](https://github.com/bpmn-io/properties-panel/blob/main/src/components/ListGroup.js#L47).

However, this might create UX problems for Errors and FormData, see [this conversation for example](https://github.com/bpmn-io/bpmn-properties-panel/pull/97#issuecomment-919809073)

__Describe the solution you'd like__

Make autoOpen a seperate parameter.

__Describe alternatives you've considered__

n/a

__Additional context__

See https://github.com/bpmn-io/bpmn-properties-panel/pull/97 as an example for a usecase

cc. @andreasgeier 
cc. @pinussilvestrus

andreasgeier:
Let's clarify this topic together. We are talking about 4 different behaviors which are currently combined in one flag.

- Do we need this behavior differentiation only on the section root or also for nested lists? – I read that values for enum in form fields are also not sorted
- Is this feasible in a short amount of time? – otherwise, we should consider alternatives

Note: `sort` and `insert on top` are scalability features while `open and focus on create` is a general usability concern for the handling of the properties panel. If it makes a significant difference for the implementation effort we could discuss if postponing scalability but keeping the usability part could be an option to support fast progress.
We also need to keep in mind that there are two opening behaviors: (1) `open the section` if it is closed on creating a new item (both, when creating the very first item but also if already items exist but the section is closed) and (2) `open the new item` that is inserted to the list.

Find [here die documentation of the current implementation](https://docs.google.com/document/d/1k2rBbbIvReLOgxaAZrUHqraaAHJh6yWqjV8FABd-Ylg/edit).

pinussilvestrus:
According to https://github.com/bpmn-io/properties-panel/issues/96#issuecomment-919892815 this one needs more clarification. Let's discuss this in an orderly manner.

MaxTru:
I added a more appropriate breakdown and will take this issue

MaxTru:
# Kick-Off on 05.10.

## Participants

* Max
* Andi
* Niklas
* Maciej

## Notes and Decisions

* Decided scope:
  * For `Lists` we will split up the existing implementation (which already has various behaviors, but they are not parameterizable), to support configuration via parameters
    * These parameters will be
      * `autoOpen`
[...]

PR #101, opened 2021-09-27
fix(list-item): auto-focus works for <select> elements

### LINKED ISSUE #100: Select Entries Can't Be Auto-Focused
__Describe the Bug__

When using the `ListItem` component you can specify which entry should be auto-focused. This doesn't work for `<select>` elements and produces an error since they don't have `#select` method.

__Steps to Reproduce__

1. Create a `ListItem` component and pass a `<select>` element's ID as the `autoFocusEntry` prop
2. Add an entry by clicking the plus
3. Error is thrown

__Expected Behavior__

Auto-focusing works for `<select>` elements.

__Environment__

- Library version: 0.4.0

philippfromme:
`<select>` elements can now be auto-focused.

RMo62gjiSP

---

Closes #100

### DIFF
--- a/src/components/ListItem.js
+++ b/src/components/ListItem.js
@@ -6,6 +6,8 @@ import {
   query as domQuery
 } from 'min-dom';
 
+import { isFunction } from 'min-dash';
+
 import CollapsibleEntry from './entries/Collapsible';
 
 /**
@@ -25,7 +27,13 @@ export default function ListItem(props) {
       const focusableInput = domQuery('.bio-properties-panel-input', entry);
 
       if (focusableInput) {
-        focusableInput.select();
+
+        if (isFunction(focusableInput.select)) {
+          focusableInput.select();
+        } else if (isFunction(focusableInput.focus)) {
+          focusableInput.focus();
+        }
+
       }
     }
   }, [ autoOpen, autoFocusEntry ]);
--- a/src/components/entries/List.js
+++ b/src/components/entries/List.js
@@ -8,6 +8,8 @@ import {
   query as domQuery
 } from 'min-dom';
 
+import { isFunction } from 'min-dash';
+
 import {
   useKeyFactory,
   usePrevious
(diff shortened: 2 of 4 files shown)


nikku:
Nica!

nikku (approved):
Very nice follow up to what could have been a UX hick-up otherwise :clap:.

/CC @andreasgeier

barmac (approved):

barmac:
Do we need any changes regarding https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/List.js#L156 ?

philippfromme:
> Do we need any changes regarding https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/List.js#L156 ?

Oh, I didn't know about that. So we have the same functionality in two places?

nikku:
> Oh, I didn't know about that. So we have the same functionality in two places?

If possible, shall move that into a single behavior (or hook).

barmac:
> > Do we need any changes regarding https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/List.js#L156 ?
> 
> Oh, I didn't know about that. So we have the same functionality in two places?

We do, though the behaviour is a bit different.

In a group list, we require to provide an ID of the entry to be focused. So theoretically, this could be a second or third entry of the item. In a nested list, we are not able to tell the ID upfront. Because of that, if you enable the autofocus, it will always be activated on the first input in an item.

pinussilvestrus:
> > Oh, I didn't know about that. So we have the same functionality in two places?
> 
> If possible, shall move that into a single behavior (or hook).

Maybe something we should consider when aligning the behaviors in general (if we want to): https://github.com/bpmn-io/properties-panel/issues/96

philippfromme:
Okay, I will fix it in both places for now. We'll tackle DRYing up that code in a later refactoring.

philippfromme:
Done. @niklas or @barmac can you have another look at this?

barmac:
Still ✅  from me.

PR #105, opened 2021-10-05
Properties panel design finishing

### LINKED ISSUE #15: Use semantic names for colors

Some of the colors in the properties panel are set via semantic names, e.g.

https://github.com/bpmn-io/properties-panel-next/blob/a0dadd1649bb248ea0ab72759e76bd905b707146/assets/properties-panel.css#L168-L174

but some are not:

https://github.com/bpmn-io/properties-panel-next/blob/a0dadd1649bb248ea0ab72759e76bd905b707146/assets/properties-panel.css#L79-L83

Let's use the first style as it allows to re-style the properties panel easier. So you don't override the properties directly:
```css
.bio-properties-panel-group {
  background-color: salmon;
}
```
but rather just change the custom property on the container level:

```css
.bio-properties-panel {
  --color-input-background: salmon
}
```

Use semantic custom properties, e.g. `--color-group-background`.

Leave it as it is, but then it will be more work to introduce re-styling in the future.

pinussilvestrus:
Related to bpmn-io/bpmn-properties-panel#90
Closes #15 

-----

Design updates for the properties panel (cf. bpmn-io/bpmn-properties-panel#90)

* update icons
* ensure titles for every action
* update the data marker dot
* add new create button for empty lists
* cleanup font + color palette
* align spacings
* ensure action hover areas match click areas
* add description for check boxes

This works alongside https://github.com/bpmn-io/bpmn-properties-panel/pull/128, which can be tested via command

```sh
npx @bpmn-io/sr bpmn-io/bpmn-properties-panel#90-ui-ux-improvements -l bpmn-io/properties-panel#ux-ui-improvements -c 'npm run start:platform'
```

image


src/components/Group.js
```diff
@@ -71,8 +72,11 @@ export default function Group(props) {
         {
           edited && <DataMarker />
         }
-        <button class="bio-properties-panel-group-header-button">
-          <GroupArrowIcon class={ open ? 'bio-properties-panel-arrow-down' : 'bio-properties-panel-arrow-right' } />
+        <button
+          title="Toggle section"
```
philippfromme, line 76: Did we agree to call this _section_ instead of _group_?

src/components/ListGroup.js
```diff
@@ -148,7 +148,10 @@ export default function ListGroup(props) {
           AddContainer
             ? (
               <AddContainer>
-                <button class="bio-properties-panel-group-header-button bio-properties-panel-add-entry">
+                <button
+                  title="Create new list item"
```
philippfromme, line 156: This title is very generic. Are we okay with that or should we make this configurable (e.g. _Create new input_)?

barmac:
Re-requesting my review so that you can continue discussion with Flip.

pinussilvestrus:
> I'm not sure what's happening but I can't see any of the titles.
> 
> cvX4w4gJZW

Are you sure you are using the correct branch? The gif does not include any of the updated styles.

src/components/Group.js
pinussilvestrus: Not sure whether we ever actively agree on a specific term here, but "section" is part of our [technical writing guide](https://github.com/camunda/camunda-modeler-design/blob/main/docs/technical-writing.md).

"Group" seems to be something we came up technically.

src/components/ListGroup.js
pinussilvestrus: It's part of our [design prototype](https://fluffy-fiesta-212e1c7c.pages.github.io/prototypes/visual-ide/v15/), so it might be ok to keep it generic. We can update it afterward if we see a need though.

/cc @andreasgeier

philippfromme: 👍🏻

PR #106, opened 2021-10-06
feat(ListGroup): improve configuration options

MaxTru:
* Place items on bottom if sorting is disabled
* Make it possible to disable autoOpening via new parameter `autoOpen`

related to #96


src/components/ListGroup.js
```diff
@@ -205,12 +209,8 @@ function createOrdering(items) {
   return items.map(i => i.id);
 }
 
-function removeDuplicates(items) {
-  return items.filter((i, index) => items.indexOf(i) === index);
-}
-
```
pinussilvestrus, line 211: Does this become obsolete? Or did it have no effect?

barmac:
It does not only change the configuration, but also the behaviour when you add new items. Note that without the proposed changes, the newly added items are stacked one on another. This means that if I add two items, the first one of these takes position no. 2. Check out this footage:

https://user-images.githubusercontent.com/28307541/136401206-244e3460-07af-436f-a7e5-029966b6524b.mp4

With this PR, only the newly added item (singular) is placed on top while the item added previously is placed as in the "sorted" order. So if add two items, the former one can be placed anywhere within the "old" items. Check out the footage:

https://user-images.githubusercontent.com/28307541/136401444-59e975e1-eea1-44ce-a4fc-b996b769597f.mp4

barmac:
At least that's how it is implemented in nested list and the current version of list group.

andreasgeier:
> So you mean it should only sort in case we close / open again?

Yes. Sort only kicks in on close/open or when switching element selection. The requirement behind this is that list items must not jump when adding or editing items.

PR #109, opened 2021-10-18
fix(ListEntry): allow custom selector for autoFocusEntry

MaxTru:
precondition for bpmn-properties-panel#133


src/components/entries/List.js
```diff
@@ -36,7 +36,7 @@ import {
  * @param {Function} [props.onRemove]
  * @param {Item[]} [props.items]
  * @param {boolean} [props.open]
- * @param {string} [props.autoFocusEntry]
+ * @param {string|boolean} [props.autoFocusEntry] either a custom selector string or true to focus the first input
```
pinussilvestrus, line 39: Can you maybe give an implementation example how an implementation will then look like (e.g. for the constraints)? Inside the [ListItems for the ListGroups](https://github.com/bpmn-io/bpmn-properties-panel/blob/main/src/provider/zeebe/properties/InputProps.js#L51) we have the specific entry id of the added item available. For the ListEntry, we define the [autoFocusEntry on the upper component](https://github.com/bpmn-io/bpmn-properties-panel/blob/main/src/provider/zeebe/properties/InputProps.js#L51).

Probably, we can get the id via the `constraints.length` to indicate the last one?

```js
autoFocusEntry={ `[data-id="${id}-constraint-${constraints.length - 1}-name"]` }
```

MaxTru: But in general one could pass any css selector really. Would you suggest to adjust the jsdoc?

pinussilvestrus (approved):
LGTM 🥇

PR #116, opened 2021-11-03
Add dropdown button

barmac:
image

This PR:

* removes default button font size which is not used anywhere (https://github.com/bpmn-io/properties-panel/commit/01adb3a25e929aca933c0d7462cb8017a4ccc459)
* adds and exposes HeaderButton
* adds and exposes DropdownButton

The components will be used in the Element Templates Group.

Related to https://github.com/bpmn-io/bpmn-properties-panel/issues/149

Test via https://github.com/bpmn-io/bpmn-properties-panel/tree/149-add-dropdown-button and `npm run start:templates`.

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -225,7 +225,6 @@
   justify-content: center;
   align-items: center;
   align-self: center;
-  font-size: var(--text-size-small);
   height: 22px;
   line-height: 22px;
   min-width: 22px;
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -201,7 +201,6 @@
   font-size: var(--text-size-base);
   height: 32px;
   user-select: none;
-  overflow: hidden;
   justify-content: space-between;
 }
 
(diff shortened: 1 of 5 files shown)


assets/properties-panel.css
```diff
@@ -778,3 +776,60 @@ textarea.bio-properties-panel-input {
...
+}
+
+.bio-properties-panel-dropdown-button__menu-item--separator {
+  width: 100%;
+  height: 1px;
+
+  padding: 0;
+  margin: 8px 0;
+
+  background-color: var(--color-grey-225-10-75);
```
pinussilvestrus, line 825: Does it make sense to use semantic color variables here as well? We do it for almost [any other color in this project](https://github.com/bpmn-io/properties-panel/blob/main/assets/properties-panel.css#L30).

barmac: I'll do that

barmac: Not sure whether I should do the same for `box-shadow` as it's quite complex.

barmac: The rest was implemented via https://github.com/bpmn-io/properties-panel/pull/116/commits/deb7684f5c298c5cf87d8135983245c8d3f8b2a3

andreasgeier:
We agreed to have a dedicated review with an artifact that contains all template implementations. Therefore, we skip design review for this PR.

src/components/DropdownButton.js
```diff
@@ -0,0 +1,108 @@
...
+    {item.entry}
+  </div>;
+}
+
+/**
+ *
+ * @param {Array<null | Element>} ignoredElements
+ * @param {Function} callback
+ */
+function useGlobalClick(ignoredElements, callback) {
```
smbea, line 90: Looks good. Just small question: why do we use this `useGlobalClick`? What does it do?

barmac: When I click outside of the dropdown button, I want the menu to close. At the same time, I don't want to close it when I click the text in the menu.

smbea: Makes sense. Thanks for the clarification

smbea (approved):
LGTM

PR #117, opened 2021-11-04
style: use semantic variables for disabled input colors

pinussilvestrus:
Little follow-up of #115. 

Let's also use semantic color variables for this.


philippfromme (approved):
Nice! 👏🏻

PR #124, opened 2021-12-09
Reexport preact && bundle library

barmac:
This PR introduces bundling to the library so that it's easier to re-export
its contents in the context of a pluggable application (Camunda Modeler).

We re-export `preact` package together with subpackages to make sure
plugin creators can depend on the same instance which is essential
for the hooks to work correctly.

Some examples of the new way (copied from the commit messages):

```javascript
import { Group, TextFieldEntry } from @bpmn-io/properties-panel;

const myGroup = {
  id: myGroup,
  component: Group,
  entries: [
    { id: customEntry, component: <TextFieldEntry /* ... */ /> }
  ]
};
```

For `preact`

```javascript
import { h } from '@bpmn-io/properties-panel/preact';

function CustomComponent(props) {
	return h('div', 'hello world!');
}
```

If you transpile from JSX, the babel `importSource` should be set
accordingly:

```json
{
  "plugins": [
    [ "@babel/plugin-transform-react-jsx", {
      "importSource": "@bpmn-io/properties-panel/preact",
      "runtime": "automatic"
    } ]
```

BREAKING CHANGES:

* Internal structure of the library is no longer exposed.
  Use root exports instead.


.babelrc.json
```diff
@@ -0,0 +1,15 @@
+{
```
barmac, line 1: Renamed so that it can be automatically parsed via `require` in the karma config.

.eslintignore
```diff
@@ -1,2 +1,3 @@
-node_modules
-lib
\ No newline at end of file
+/node_modules
+/lib
```
barmac, line 2: Prepending slash means that the ignore will be applied only to the root of the repo.

nikku (approved):
A fundamental improvement. :+1:

PR #128, opened 2022-01-07
fix: rollup now sets up correct paths for preact on windows builds

### LINKED ISSUE #127: Builds failing on Windows due to pathing issues
__Describe the Bug__

On Windows specifically, the package cannot be built at all, as multiple modules relating to preact cannot be resolved.

```
ERROR in ./preact/hooks/dist/hooks.module.js 1:0-37
Module not found: Error: Can't resolve '....' in 'C:\dev\bpmn-io\properties-panel\properties-panel\preact\hooks\dist'
 @ ./src/PropertiesPanel.js 1:0-54 84:30-38 85:2-11
 @ ./test/spec/PropertiesPanel.spec.js 5:0-50 258:21-36
 @ ./test/ sync .spec\.js$ ./spec/PropertiesPanel.spec.js test/spec/PropertiesPanel.spec.js
 @ ./test/testBundle.js 1:17-57
 ```

__Steps to Reproduce__

1. Clone the repo on a Windows machine
2. Execute `npm install`
3. Execute `npm run dev`

__Expected Behavior__

We expect the build to complete without errors.

__Environment__

- Host: Node v14
- OS: Window 10
- Library version: 0.10.0

### PR DESCRIPTION
**Pull request by Skaiir**
Closes #127.

This also includes changes to the CI config, now triggering builds on windows, mac and linux platforms.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,4 +1,4 @@
-import { parse as parsePath, relative as relativePath } from 'path';
+import path, { parse as parsePath, relative as relativePath } from 'path';
 import { replaceInFile } from 'replace-in-file';
 
 import babel from '@rollup/plugin-babel';
@@ -64,9 +64,8 @@ function rewirePreactSubpackages() {
                 filePath = args.pop();
 
           const { dir } = parsePath(filePath);
-          const pathToPreact = relativePath(dir, './preact');
-
-          return `${importGroup}${pathToPreact}${afterImport}`;
+          const posixPathToPreact = relativePath(dir, './preact').split(path.sep).join(path.posix.sep);
+          return `${importGroup}${posixPathToPreact}${afterImport}`;
         }
       });
     }
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -5,7 +5,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ ubuntu-latest ]
+        os: [ macos-latest, ubuntu-latest, windows-latest ]
         node-version: [ 14 ]
 
     runs-on: ${{ matrix.os }}
@@ -29,8 +29,16 @@ jobs:
     - name: Install dependencies
       run: npm ci
     - name: Build
+      if: matrix.os == 'ubuntu-latest'
       env:
         COVERAGE: 1
+        TEST_BROWSERS: ChromeHeadless
+      run: npm run all
+    - name: Build
+      if: matrix.os != 'ubuntu-latest'
+      env:
+        TEST_BROWSERS: ChromeHeadless
       run: npm run all
     - name: Upload coverage
+      if: matrix.os == 'ubuntu-latest'
       run: npx codecov

### CODE REVIEW

**Review by MaxTru — COMMENTED** [review]
LGTM!

Will not approve yet, so to give @philippfromme as chance to review (being a windows user)

File: rollup.config.js [hunk 1]
```diff
@@ -65,8 +65,9 @@ function rewirePreactSubpackages() {
 
           const { dir } = parsePath(filePath);
           const pathToPreact = relativePath(dir, './preact');
+          const escapedPathToPreact = './' + pathToPreact.replace('\\', '\\\\');
```
philippfromme, line 68: Maybe something like

```
import { sep as pathSeparator } from 'path';

// ...

const pathToPreact = relativePath(dir, './preact').split(pathSeparator).join('/');

return `${ importGroup }./${ pathToPreact }${ afterImport }`;
```

would be a bit more elegant as it would produce something like `./../..` instead of `./..\\..`

philippfromme (commented):
👍🏻

rollup.config.js
nikku: Stackoverflow agrees: https://stackoverflow.com/questions/53799385/how-can-i-convert-a-windows-path-to-posix-path-using-node-path.

Skaiir: There's also a way to do it via the path/posix sub-package I'm pretty sure, but it's only clean if you use the require() way of importing so agreed I'll use that.

rollup.config.js
```diff
@@ -64,9 +64,8 @@ function rewirePreactSubpackages() {
                 filePath = args.pop();
 
           const { dir } = parsePath(filePath);
-          const pathToPreact = relativePath(dir, './preact');
-
-          return `${importGroup}${pathToPreact}${afterImport}`;
+          const posixPathToPreact = relativePath(dir, './preact').replace(path.sep, path.posix.sep);
```
philippfromme, line 67: The `'./'` prefix is now gone. On purpose?

Skaiir: Yes, as it turns out that prefix was only needed because of the backslash. ..\.. just wasn't interpreted as a proper path, while ./..\.. was. But we never really needed it in the first place.

Skaiir: ../.. though works absolutely fine

nikku: Please be mindful of a replace that does not do what you would expect: 

image

I know this is necessary in this place, still, to properly replace all occurences you gotta do `a.split(sep).join(otherSep)`.

Skaiir: Ah yes I blame my history with C# here. 

I've made the adjustments and squashed the commits. :)

philippfromme:
Last thing: Can we squash the commits so we only have 2 (one for the fix and one for the additional builds)?

Skaiir:
> Last thing: Can we squash the commits so we only have 2 (one for the fix and one for the additional builds)?

Tis done.

philippfromme (approved):
I've updated the commit messages. Generally, the prefix is always lowercase.

Great fix! 👏🏻

nikku:
:tada:

PR #129, opened 2022-01-07
fix(keyboard navigation): Delete button now shows on tabbing

Skaiir:
Related to [#505](https://github.com/bpmn-io/bpmn-js-properties-panel/issues/505)

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -642,7 +642,9 @@ textarea.bio-properties-panel-input {
 }
 
 .bio-properties-panel-list-entry-item:hover .bio-properties-panel-remove-list-entry,
-.bio-properties-panel-collapsible-entry:hover > .bio-properties-panel-collapsible-entry-header .bio-properties-panel-remove-entry {
+.bio-properties-panel-collapsible-entry:hover > .bio-properties-panel-collapsible-entry-header .bio-properties-panel-remove-entry,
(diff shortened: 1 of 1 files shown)


Skaiir:
This does not yet build on windows and mac, as [#505](https://github.com/bpmn-io/properties-panel/pull/128) hasn't yet been merged. I've however tested the changes against the new CI, and it's working as intended.

pinussilvestrus:
General hint: make sure you always reference the fixed issue in your commit message, this way we can later check out what a commit was targeting at

```text
fix(keyboard navigation): Delete button now shows on tabbing

Closes #505
```

Skaiir:
Ah right so not only the Branch, makes sense. Should I somehow modify the commit message now? @pinussilvestrus

pinussilvestrus:
Yeah, let us clean up the commits before merging. If you don't already know, via the command line, you can simply use `git commit --amend` to update the message.

pinussilvestrus:
By the way, we should add `Related to` and not `Closes` because the issue is not in this project, but the PR is fixing something in another, upstream repo. We also usually add the `fixed upstream` label to this issue then afterwards and wait for a release.

pinussilvestrus (approved):
Very nice, works for me 👍 

Kapture 2022-01-07 at 16 01 38

PR #131, opened 2022-01-17
wip: generate type declarations

### LINKED ISSUE #65: Add type linting + generation
__What should we do?__

Currently, we only have very limited type support for some parts of the library (cf. https://github.com/bpmn-io/properties-panel-next/pull/61/commits/ffa26d295528903166145aac10ec9d685f45e1ec). We could follow the same route as we do in [form-js](https://github.com/bpmn-io/form-js).

__Why should we do it?__

Better development support + discoverability of this library.

barmac:
This adds TypeScript declarations to the package.

Closes #65

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8351,6 +8351,12 @@
         "mime-types": "~2.1.24"
       }
     },
+    "typescript": {
+      "version": "4.5.4",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.5.4.tgz",
+      "integrity": "sha512-VgYs2A2QIRuGphtzFV7aQJduJ2gyfTljngLzjpfW9FoYZF6xuw1W0vW9ghCKLfcWrCFxK81CSGRAvS1pn4fIUg==",
+      "dev": true
+    },
     "ua-parser-js": {
       "version": "0.7.28",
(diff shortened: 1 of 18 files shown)


barmac:
Basic infrastructure is in place and now I will focus on fixing the errors.

barmac:
I've fixed the types, but I believe we should also cross-test this with `bpmn-js-properties-panel` or have an integration-test app similar to [the one in form-js](https://github.com/bpmn-io/form-js/tree/master/packages/form-js-integration) before we merge this.

barmac:
I created a branch which implements the type linting in `bpmn-js-properties-panel` but it's far from being finished. Check out the link: https://github.com/bpmn-io/bpmn-js-properties-panel/tree/add-type-linting

barmac:
Although the type linting results are all green, the hundreds of errors within the integration test branch linked above makes me believe we don't sufficiently test the declarations. Also, it's not that easy to handle the broken JSDoc typings we have in bpmn-js-properties-panel. This means I don't expect that we quickly adopt type linting in that project and thus cannot verify the typings here.

Consequently, I am closing this PR. I believe we should start with limited, public API-only declarations written by hand. Regarding the code in this PR, there are some fixes which we can use (e.g. `spellCheck` fix) in a separate contribution.

PR #132, opened 2022-01-20
Add basic accessibility tests

pinussilvestrus:
This
* adds basic a11y testing infrastructure via `axe-core`
* defines basic rules set we testing against 
  * `best-practice`
  * `wcag2a`
  * `wcag2aa`,
  * `cat.semantics` (some additional rules to ensure we use semantic HTML, something we focus on in the past)
* adds unit tests for all components 
* fixes a minor a11y issue with the `Simple` component (no label)

### DIFF
--- a/karma.config.js
+++ b/karma.config.js
@@ -107,7 +107,10 @@ module.exports = function(karma) {
         modules: [
           'node_modules',
           absoluteBasePath
-        ]
+        ],
+        fallback: {
+          'crypto': false
+        }
       },
       devtool: 'eval-source-map'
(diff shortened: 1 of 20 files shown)


pinussilvestrus:
Disclaimer: as already discussed in the [weekly lightning talk](https://github.com/bpmn-io/internal-docs/issues/423), this only ensures basic rules, but does not guarantee our components are accessible.

philippfromme (approved):
This is good stuff! Well done. 👏🏻

PR #134, opened 2022-02-21
feat: render components instead of elements

philippfromme:
* all group and entry components specified as `component` are actual components, not elements
* props can be passed from parent to child components
* props specified when creating group or entry will be passed to component

---

Required by https://github.com/bpmn-io/bpmn-js-properties-panel/pull/590

### DIFF
--- a/src/components/entries/Collapsible.js
+++ b/src/components/entries/Collapsible.js
@@ -12,11 +12,12 @@ import {
 
 export default function CollapsibleEntry(props) {
   const {
-    id,
+    element,
     entries = [],
+    id,
     label,
-    remove,
-    open: shouldOpen
+    open: shouldOpen,
+    remove
   } = props;
 
   const [ open, setOpen ] = useState(shouldOpen);
@@ -64,7 +65,17 @@ export default function CollapsibleEntry(props) {
         open ? 'open' : ''
       ) }>
         {
-          entries.map(e => e.component)
+          entries.map(entry => {
+            const {
+              component: Component,
+              id
+            } = entry;
+
+            return <Component
+              { ...entry }
+              key={ id }
+              element={ element } />;
+          })
         }
       </div>
     </div>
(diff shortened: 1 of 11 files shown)


src/components/entries/Collapsible.js
```diff
@@ -64,7 +65,17 @@ export default function CollapsibleEntry(props) {
...
+            const {
+              component: Component,
+              id
+            } = entry;
+
+            return <Component
+              { ...entry }
+              key={ id }
+              element={ element } />;
+          })
```
pinussilvestrus, line 78: With this one, we could actually get rid of setting the id twice (on the entry, on the component), by defaulting the component id with the entry id.

Example: https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/provider/bpmn/properties/IdProps.js#L65

pinussilvestrus: It's not needed, but just something that came to my mind.

philippfromme: True, actually. Let me have a look and see how much effort that would be.

philippfromme: @pinussilvestrus I've looked into it and we actually have to set different IDs for the entry and the actual DOM element in the future (e.g. https://github.com/bpmn-io/bpmn-js-properties-panel/blob/spike-linting/src/provider/zeebe/properties/TaskDefinitionProps.js#L133). The IDs of entries have to be stable (e.g. `task-definition-type`) while the IDs of the DOM elements have to be dynamic (e.g. `extensionElements.0.type`) for them to be addressable. So let's keep it as it is.

philippfromme:
@pinussilvestrus @barmac The sooner we merge these pull requests fewer merge conflicts we will have to resolve. 😉

barmac:
I'll look into this right after the topic I am on now. So before noon.

barmac:
LGTM

philippfromme:
> LGTM

Can I haz approval? 🍔

barmac (approved):
Yes

barmac:
Hmm I just noticed that there's no _Breaking changes_ part in the commit. This PR is a breaking change though.

PR #137, opened 2022-03-09
Be Able to Show Entry and Error through Event

philippfromme:
## Events

`EventContext` is introduced. Through this context, one can subscribe to events using the `useEvent` hook.

```js
function Foo() {
  useEvent('foo', (event) => console.log(event));

  // ...
}
```

## Event Buffering

The `useEventBuffer` hook is introduced to buffer events that will be re-fired during the passive effect phase. If a component is rendered after an event was fired it will still catch the event.

```js
function Foo() {
  useEvent('foo', (event) => console.log(event));

  // ...
}

function Bar() {
  useEventBuffer([ 'foo' ], eventBus);

  return <Foo />;
}
```

It allows the user to do

```js
selection.select(element);

eventBus.fire('propertiesPanel.showEntry', { ... });
```

without using `setTimeout` or similar workarounds.

## Showing Entries

The `useShowEntryEvent` hook is introduced to make subscribing to `propertiesPanel.showEntry` events easy. It expects a `show` callback that will be called whenever a `propertiesPanel.showEntry` event is caught. If it returns true the entry is shown. The `ref` returned by the hook can be used to focus an input whenever an entry is shown.

```js
function Foo() {
  const show = useCallback(({ foo }) => foo === 'foo', []);

  const ref = useShowEntryEvent(show);

  return <input type="text" ref={ ref } />;
}
```

## Showing Errors

The `useShowEntryEvent` hook is introduced to make subscribing to `propertiesPanel.showError` events easy. It also expects a `show` callback and an optional second argument `inputs`. Whenever any of the values in `inputs` changes the temporary error will be reset.

```js
function Foo() {
  const [ value, setValue ] = useState('foo');

  const show = useCallback(({ foo }) => foo === 'foo', [ 'foo', 'bar' ]);

  const temporaryError = useShowErrorEvent(show, [ value ]);

  // ...
}
```

## Showing Parent Groups When Showing Entries

For an entry to be shown all of its parent groups must be shown as well. An `onShow` callback is introduced that can be called by a child component whenever it's shown. The parent component will then be shown as well. The `onShow` callback is provided through the closest context provider of the newly introduced `PropertiesPanelContext`. Providing the callback through a context instead of a prop means that it doesn't have to be passed explicitly when creating a custom entry.

```js
import { Group, TextFieldEntry } from '@bpmn-io/properties-panel';

function FooEntry() {

  // TextFieldEntry component will get onShow through context provider of its parent Group component
  return <TextFieldEntry />;
}

const group = {
  component: Group,
  entries: [
    {
      component: FooEntry
    }
  ]
};
```

---

Required by https://github.com/bpmn-io/bpmn-js-properties-panel/pull/601

### DIFF
--- a/src/components/entries/TextField.js
+++ b/src/components/entries/TextField.js
@@ -9,10 +9,15 @@ import {
 
 import classnames from 'classnames';
 
+import { isFunction } from 'min-dash';
+
 import {
-  usePrevious
+  usePrevious,
+  useShowEntryEvent,
+  useShowErrorEvent
 } from '../../hooks';
 
+const noop = () => {};
 
 function Textfield(props) {
 
@@ -23,8 +28,11 @@ function Textfield(props) {
     label,
     onInput,
     feel = false,
-    value = ''
+    value = '',
+    show = noop
   } = props;
+
+  const ref = useShowEntryEvent(show);
 
   const handleInput = useMemo(() => {
     return debounce(({ target }) => onInput(target.value.length ? target.value : undefined));
@@ -37,6 +45,7 @@ function Textfield(props) {
         {feel && <FeelIcon feel={ feel } label={ label } />}
       </label>
       <input
+        ref={ ref }
         id={ prefixId(id) }
         type="text"
         name={ id }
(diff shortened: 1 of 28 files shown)


src/context/EventContext.js
```diff
@@ -0,0 +1,40 @@
+/**
```
pinussilvestrus, line 1: Pretty nice with those examples 👏

src/components/entries/TextField.js
```diff
@@ -23,9 +28,12 @@ function Textfield(props) {
     label,
     onInput,
     feel = false,
-    value = ''
+    value = '',
+    show = noop
```
pinussilvestrus, line 32: So every entry (e.g. property entry) has to define on its own whether it shall be shown or not? What would such a callback look like? Given the [referenced branch in bpmn-js-properties-panel](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/601/files#diff-7650658698a45a7549c8cdba125ffa207d194a6be03101347fb1bf88b62d6b5aR138) I see that this could mean quite an effort. Will there be an easier way to provide such a callback?

pinussilvestrus (commented):
Pretty pretty nice stuff 👏  I just have one question regarding the `show` callback. Besides, it looks pretty solid. The examples and summary helped a lot 👍 

Looking forward to seeing this stuff in action 🙂

src/components/entries/TextField.js
philippfromme: First of all, providing a `show` callback is optional so not every entry has to implement one. The generic callback is necessary because it cannot be assumed e.g. that every entry is shown based on an ID or a path. To make things easier we can create utilities in bpmn-js-properties-panel to make creating `show` callbacks less repetitive (e.g. `function showPath(id, path) { ... }`).

pinussilvestrus: >  To make things easier we can create utilities in bpmn-js-properties-panel to make creating show callbacks less repetitive (e.g. function showPath(id, path) { ... }).

👍

pinussilvestrus (approved):
LGTM 👏

philippfromme:
Thanks @pinussilvestrus, will follow up with https://github.com/bpmn-io/bpmn-js-properties-panel/pull/601/commits shortly!

PR #139, opened 2022-03-22
feat(group): make inital open state configurable

marstamm:
related to https://github.com/bpmn-io/bpmn-js-properties-panel/issues/618

for integration PR in bpmn-js properties panel, see https://github.com/bpmn-io/bpmn-js-properties-panel/pull/621

### DIFF
--- a/src/components/Group.js
+++ b/src/components/Group.js
@@ -27,12 +27,13 @@ export default function Group(props) {
     element,
     entries = [],
     id,
-    label
+    label,
+    shouldOpen = false,
   } = props;
 
   const [ open, setOpen ] = useLayoutState(
     [ 'groups', id, 'open' ],
-    false
+    shouldOpen
   );
   const toggleOpen = () => setOpen(!open);
 
--- a/src/context/LayoutContext.js
+++ b/src/context/LayoutContext.js
@@ -5,7 +5,7 @@ import {
 const LayoutContext = createContext({
   layout: {},
   setLayout: () => {},
-  getLayoutForKey: () => {},
+  getLayoutForKey: (key, defaultValue) => defaultValue,
   setLayoutForKey: () => {}
 });
 
--- a/src/context/LayoutContext.js
+++ b/src/context/LayoutContext.js
@@ -5,7 +5,7 @@ import {
 const LayoutContext = createContext({
   layout: {},
   setLayout: () => {},
-  getLayoutForKey: (key, defaultValue) => defaultValue,
+  getLayoutForKey: () => {},
   setLayoutForKey: () => {}
 });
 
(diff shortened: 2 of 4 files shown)


src/components/Group.js
```diff
@@ -27,12 +27,13 @@ export default function Group(props) {
     element,
     entries = [],
     id,
-    label
+    label,
+    initiallyOpen = false,
```
pinussilvestrus, line 31: Why not `shouldOpen` as we have in [other components](https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/Collapsible.js#L25)?

pinussilvestrus: We should also make sure to update our [type docs](https://github.com/bpmn-io/properties-panel/blob/main/src/PropertiesPanel.js#L66) (although we don't validate with them by now).

marstamm:
Ready for review. I also adjusted the downstream integration: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/621

src/context/LayoutContext.js
```diff
@@ -5,7 +5,7 @@ import {
 const LayoutContext = createContext({
   layout: {},
   setLayout: () => {},
-  getLayoutForKey: () => {},
+  getLayoutForKey: (key, defaultValue) => defaultValue,
```
pinussilvestrus, line 8: Do we need to be that explicit at this stage?

pinussilvestrus (commented):
Looks good in general, makes sense to make it possible to set the initial state from a parent component. 👍 

Just for my understanding: did we consider building it via the `useEvent` API [we recently introduced](https://github.com/bpmn-io/properties-panel/pull/137)? So to make it also possible to open groups from elsewhere? 

Probably that's a different use case because this `open` state should also be persisted, right?

src/context/LayoutContext.js
marstamm: You are right, we can mock it in the tests instead. I went ahead and did that

marstamm:
> Just for my understanding: did we consider building it via the useEvent API https://github.com/bpmn-io/properties-panel/pull/137? So to make it also possible to open groups from elsewhere?

I did consider going this way. However, the use case is slightly different and not as clean as solving it via props. E.g., we would always have to fire the event right after rendering while also checking that the state is persisted. Also, the current API only supports elements and not groups.

In the end, I decided to go with `shouldOpen`, as we also have those 2 options side-by-side in the ListGroup component:
https://github.com/bpmn-io/properties-panel/blob/00c4b3be7c83927c77758d4ab6c2e58400e97436/src/components/ListGroup.js#L41

nikku (commented):
I like the way this is going + agree to https://github.com/bpmn-io/properties-panel/pull/139#issuecomment-1077367615.

@pinussilvestrus Please give your :+1: to confirm implementation.

pinussilvestrus (approved):
Thanks for the clarification 👏  I agree 🙂

PR #145, opened 2022-04-06
fix: remove unnecessary properties panel padding

Skaiir:
Related to https://github.com/camunda/camunda-modeler/issues/2831.

The negative margin is there to 'hide' the last group's separator when a scrollbar is visible.

This avoids that:

image

This has the side effect of moving the scrollbar itself down a little. Not really visible, and definitely a little weird but I've tried multiple other css tricks without success.

How it will look:

image

Worth testing on non windows platforms if the scrollbars look different.

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -197,7 +197,6 @@
   overflow-y: auto;
   overflow-x: hidden;
   flex: 1;
-  padding-bottom: 70px;
 }
 
 /**


andreasgeier:
@Skaiir pls provide an artifact for design review.

barmac (changes requested):
Check out the comment.

Skaiir:
@andreasgeier We're going to first remove the padding from the props panel, and then add the margin inside of the modeler when I integrate the new version. So we'll defer design review to a little later if that sounds correct to you.

barmac (approved):

PR #150, opened 2022-05-16
fix: prevent debounced input from losing value when focused

### LINKED ISSUE #146: Debounced Input Sometimes Loses Input
__Describe the Bug__

 Due to the debounced input (cf. https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/TextField.js#L38) sometimes the input is lost. As the component is controlled after the debounce delay the value is passed back to the input as a prop and will override any input that happened inbetween.

__Steps to Reproduce__

1. Focus a `TextField` input
2. Type rather fast
3. Sometimes the input is lost

__Expected Behavior__

No input is ever lost while the input is focused.

__Environment__

- Library version: 0.13.1

philippfromme:
Prevents text input from losing value due to prop changes while focused.

Closes #146

### DIFF
--- a/src/components/entries/TextField.js
+++ b/src/components/entries/TextField.js
@@ -12,6 +12,7 @@ import classnames from 'classnames';
 import { isFunction } from 'min-dash';
 
 import {
+  useLocalValue,
   usePrevious,
   useShowEntryEvent,
   useShowErrorEvent
@@ -34,9 +35,16 @@ function Textfield(props) {
 
   const ref = useShowEntryEvent(show);
 
-  const handleInput = useMemo(() => {
-    return debounce(({ target }) => onInput(target.value.length ? target.value : undefined));
-  }, [ onInput, debounce ]);
+  const debouncedOnInput = useMemo(() => {
+    return debounce((value) => onInput(value.length ? value : undefined));
+  }, [ debounce, onInput ]);
+
+  const [
+    localValue,
+    localOnInput,
+    onFocus,
+    onBlur
+  ] = useLocalValue(value, debouncedOnInput);
 
   return (
     <div class="bio-properties-panel-textfield">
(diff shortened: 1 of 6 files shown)


src/components/entries/TextField.js
```diff
@@ -53,10 +56,10 @@ function Textfield(props) {
         autoComplete="off"
         disabled={ disabled }
         class="bio-properties-panel-input"
-        onInput={ handleInput }
-        onFocus={ props.onFocus }
-        onBlur={ props.onBlur }
```
pinussilvestrus, line 58: Good that we removed these props; we never used them (in the entry components). 👍

nikku:
So what does this do, and why? :runner:

philippfromme:
@nikku It doesn't accept value via prop changes while focused as there is no good reason to do so.

pinussilvestrus:
> @nikku It doesn't accept value via prop changes while focused as there is no good reason to do so.

That makes sense to me. When I'm typing (I have the focus), I do not want to get overruled by random prop changes.

nikku:
> It doesn't accept value via prop changes while focused as there is no good reason to do so.

Aha!

Still wondering why Preact does not handle this out of the box. It has all information it needs, and applying the diff (no-op) shall keep the cursor. Did you find an answer for that? What you've built looks like a workaround for what the framework shall do for us.

nikku:
> When I'm typing (I have the focus), I do not want to get overruled by random prop changes.

Does not work with undo though. :see_no_evil:

philippfromme:
> Does not work with undo though. 🙈

😭

philippfromme:
This is what you get when you disable native undo of an input.

nikku:
Happy to re-enable. I mean, one day we'll have an append only log anyway. Always do, never look back!

philippfromme:
Question is does the user expect undo in an input to actually trigger undo of our command stack. 🤔

philippfromme:
Will keep digging for a solution.

philippfromme:
Closed in favor of https://github.com/bpmn-io/properties-panel/pull/154.

PR #151, opened 2022-05-18
Sticky headers

MaxTru:
closes camunda/camunda-modeler-design#118

:warning: Please notice this does not yet include the `Scroll to focus on new added list item` => I will tackle this seperately as a follow-up.

Try this out yourself with:
```
npx @bpmn-io/sr "bpmn-io/bpmn-js-properties-panel" -l "bpmn-io/properties-panel#sticky-headers"  
```

StickyHeader

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -218,6 +218,7 @@
   height: 32px;
   user-select: none;
   justify-content: space-between;
+  margin-bottom: -1px; /* avoid double borders */
   position: relative;  /* browsers not supporting sticky */
   position: -webkit-sticky;  /* for safari */
   position: sticky;
@@ -239,7 +240,6 @@
 .bio-properties-panel-group-header.sticky {
   background-color: var(--sticky-group-background-color);
   border-bottom: 1px solid var(--sticky-group-bottom-border-color);
-  margin-bottom: -1px;
 }
 
 .bio-properties-panel-group-header-buttons {
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -45,6 +45,9 @@
   --group-background-color: var(--color-white);
   --group-bottom-border-color: var(--color-grey-225-10-75);
 
+  --sticky-group-background-color: var(--color-grey-225-10-95);
+  --sticky-group-bottom-border-color: var(--color-grey-225-10-75);
+
   --add-entry-fill-color: var(--color-grey-225-10-35);
   --add-entry-hover-fill-color: var(--color-white);
   --add-entry-hover-background-color: var(--color-blue-205-100-50);
@@ -144,7 +147,6 @@
   align-items: center;
   font-size: var(--text-size-base);
   padding: 16px 10px;
-  margin-bottom: 2px;
   background-color: var(--header-background-color);
   border-bottom: 1px solid var(--header-bottom-border-color);
   width: 100%;
@@ -205,10 +207,10 @@
 .bio-properties-panel-group {
   background-color: var(--group-background-color);
   border-bottom: 1px solid var(--group-bottom-border-color);
+  position: relative;
 }
 
 .bio-properties-panel-group-header {
-  position: relative;
   display: flex;
   flex-direction: row;
   align-items: center;
@@ -216,6 +218,10 @@
   height: 32px;
   user-select: none;
   justify-content: space-between;
+  position: -webkit-sticky;  /* for safari */
+  position: sticky;
+  top: 0;
+  z-index: 1;
 }
 
 .bio-properties-panel-group-header .bio-properties-panel-group-header-title {
@@ -227,6 +233,12 @@
 
 .bio-properties-panel-group-header.open .bio-properties-panel-group-header-title {
   font-weight: 500;
+}
+
+.bio-properties-panel-group-header.sticky {
+  background-color: var(--sticky-group-background-color);
+  border-bottom: 1px solid var(--sticky-group-bottom-border-color);
+  margin-bottom: -1px;
 }
 
 .bio-properties-panel-group-header-buttons {
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -218,6 +218,7 @@
   height: 32px;
   user-select: none;
   justify-content: space-between;
+  position: relative;  /* browsers not supporting sticky */
   position: -webkit-sticky;  /* for safari */
   position: sticky;
   top: 0;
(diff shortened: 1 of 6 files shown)


pinussilvestrus:
I like that it also works via keyboard ❤️ Although it looks a bit shaky (on my Windows machine).

https://user-images.githubusercontent.com/9433996/169143889-a2fdb680-c01f-41cb-983c-197b32da1da7.mp4

src/hooks/useStickyIntersectionObserver.js
```diff
@@ -0,0 +1,55 @@
...
+ * @param {setSticky} setSticky
+ */
+export function useStickyIntersectionObserver(ref, scrollContainerSelector, setSticky) {
+  useEffect(() => {
+    let observer;
+
+    if (ref.current) {
+      const scrollContainer = domQuery(scrollContainerSelector);
+
+      observer = new IntersectionObserver((entries) => {
```
pinussilvestrus, line 31: IE has[ no support](https://caniuse.com/?search=IntersectionObserver) for this, but I think we are fine with that, right?

MaxTru: Yes, that would be my understanding. 

@nikku and @barmac also discussed with me that we can use `position: sticky`, which is also not supported by IE.

test/spec/hooks/useStickyIntersectionObserver.spec.js
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+});
+
+
+// helpers ////////////////////
+
+function noop() {}
+
+function mockIntersectionObserver(props) {
```
pinussilvestrus, line 90: 👍

pinussilvestrus: Testing observers is really difficult, so I'm happy we find a way to mock it (reliably). Context: [Resize observers in the Camunda Modeler.](https://github.com/camunda/camunda-modeler/blob/develop/client/src/app/primitives/__tests__/TabLinksSpec.js#L397)

barmac:
Does `position: sticky` not work for this problem? What is the reason we re-implement it with the Intersection Observer?

I'd suggest to think about this in the context of [the progressive enhancement philosophy](https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement). In short words: We make sure that the tool works for as many users as possible, but also make use of new features if they are supported by the browser. This is contrary to:

* We ship tool only for the newest browsers and don't care if it breaks in a slightly outdated browser.
* We ship tool with exactly the same experience even if it requires us to not use new features at all or to ship significantly more code.

Applied to the subject of sticky headers, I think we should try to make use of `position: sticky` and ship the old solution for browsers which don't support it. This could look like snippet below:

```css
.group {
  position: relative; /* old browsers */
  position: sticky;
}
```

Let me know what you think.

barmac:
Also, notice that `new IntersectionObserver` will throw an error in IE11 unless it's polyfilled while unrecognized CSS degrades gracefully.

barmac:
BTW we already use `position: sticky` in the decision table:
image

pinussilvestrus:
Correct me if I'm wrong, [but we already make use of `position: sticky`, right?](https://github.com/bpmn-io/properties-panel/pull/151/files#diff-398d379cec6e774a099d60df565c9d947f4799b2c1bfab8825c3ba4fd0d900faR221-R222) The observer only deals with changing the "mode to provide more style changes when it reaches the sticky state.

> ship the old solution for browsers which don't support it.

👍 Make sure we do not blow up

barmac:
Oh OK. I didn't notice because I was too focused on the intersection observer 😅

barmac:
In that case, most of what I wrote is irrelevant. Sorry for possible confusion.

I found a flickering problem which is triggered due to the negative margin:


https://user-images.githubusercontent.com/28307541/169269270-9af123c5-2f38-4986-a2e2-7b754c8dbd87.mov

assets/properties-panel.css
```diff
@@ -205,17 +207,22 @@
...
   flex-direction: row;
   align-items: center;
   font-size: var(--text-size-base);
   height: 32px;
   user-select: none;
   justify-content: space-between;
+  position: relative;  /* browsers not supporting sticky */
+  position: -webkit-sticky;  /* for safari */
+  position: sticky;
+  top: 0px;
```
barmac, line 224:
```suggestion
  top: 0;
```

No need to use unit for zero :)

MaxTru: :+1:

src/hooks/useStickyIntersectionObserver.js
```diff
@@ -0,0 +1,56 @@
...
+          setSticky(true);
+        }
+        else if (entries[0].intersectionRatio === 1) {
+          setSticky(false);
+        }
+      },
+      {
+        root: scrollContainer,
+        rootMargin: '2px 0px 999999% 0px', // Threshold to avoid flickering | Use bottom margin to avoid stickyness when scrolling out to bottom
+        threshold: [ 1 ]
```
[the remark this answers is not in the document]
barmac: Ok I tried out delta but it does not help. The problem is caused by changing size of the element.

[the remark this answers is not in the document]
MaxTru: I am puzzled. Maybe we accept the double border effect. Or maybe @andreasgeier has a CSS idea.

MaxTru:
If visually acceptable (cc @andreasgeier ), we could fix it with this:
https://github.com/bpmn-io/properties-panel/pull/151/commits/010ef1b0f5afca03a9674db8c9af77f064e9ce2f

Now we should not have the flickering

barmac (approved):
LGTM

barmac:
Maybe one more thing regarding the IntersectionObserver. I'd suggest to make it optional, so that the properties panel does not blow up because of this minor feature if it's opened in an old browser.

barmac:
> Maybe one more thing regarding the IntersectionObserver. I'd suggest to make it optional, so that the properties panel does not blow up because of this minor feature if it's opened in an old browser.

Added via https://github.com/bpmn-io/properties-panel/pull/151/commits/3b8165cd0ecde1dd0f6b70a1a65d807e6e51bdb4

✅  from me with this change

barmac (approved):

MaxTru:
> > Maybe one more thing regarding the IntersectionObserver. I'd suggest to make it optional, so that the properties panel does not blow up because of this minor feature if it's opened in an old browser.
> 
> Added via [3b8165c](https://github.com/bpmn-io/properties-panel/commit/3b8165cd0ecde1dd0f6b70a1a65d807e6e51bdb4)
> 
> white_check_mark from me with this change

Great addition, thanks

PR #152, opened 2022-05-19
Enable generic multiple and empty state

### LINKED ISSUE #69: Enable multi-select (and empty) state for the properties panel
__What should we do?__

We need to consider the multi-select scenario for the properties panel. Currently, the panel only works correctly if one element is select at a time.

(Optional in the Zeebe scope:) We should also consider the no-selection / empty state. This is not possible in the BPMN and DMN world but we have an empty state in Forms.

Cf. the [visual IDE prototype](https://fluffy-fiesta-212e1c7c.pages.github.io/prototypes/visual-ide/v17/) (Select "empty state" and "multi state" elements) 

image

image

For both cases: background color is `--color-grey-225-10-95`

__Why should we do it?__

There is no current solution. In a multi-select scenario, the panel shows just one of the select elements. It is even possible to edit this (random?) element in the panel.

Related to https://github.com/camunda/camunda-modeler-design/issues/100

pinussilvestrus:
Closes #69 

* Improve placeholder to provide multiple and empty states
* Allows integrators (e.g. bpmn-js-properties-panel) to provide contents 

Note (/cc @andreasgeier): I used `--color-grey-225-10-35` for the text color as the color [in the prototype ](https://fluffy-fiesta-212e1c7c.pages.github.io/prototypes/visual-ide/v18/) offers too less contrast (`--color-grey-225-10-55`). It even broke the built-in a11y integration test (which is a good sign our tests work 👍 )

Example

```sh
npx @bpmn-io/sr "bpmn-io/bpmn-js-properties-panel#use-updated-placeholder" -l "bpmn-io/properties-panel#69-multi-and-empty-state" -c "npm run start:cloud" 
```

**Multiple state**

image

**Empty state**

Only in theory, not possible via UI.

image

### DIFF
--- a/src/components/Placeholder.js
+++ b/src/components/Placeholder.js
@@ -1,0 +1,23 @@
+/**
+ * @typedef { {
+ *  text: (element: object) => string,
+ *  icon?: (element: Object) => import('preact').Component
+ * } } PlaceholderDefinition
+ *
+ * @param { PlaceholderDefinition } props
+ */
+export default function Placeholder(props) {
+  const {
+    text,
+    icon: Icon
+  } = props;
+
+  return (
+    <div class="bio-properties-panel open">
+      <section class="bio-properties-panel-placeholder">
+        { Icon && <Icon class="bio-properties-panel-placeholder-icon" /> }
+        <p class="bio-properties-panel-placeholder-text">{ text }</p>
+      </section>
+    </div>
+  );
+}
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -36,7 +36,8 @@
   --description-code-border-color: var(--color-grey-225-10-85);
   --description-list-item-color: var(--color-grey-225-10-35);
 
-  --placeholder-color: var(--color-grey-225-10-75);
+  --placeholder-color: var(--color-grey-225-10-35);
+  --placeholder-background-color: var(--color-grey-225-10-95);
 
   --header-background-color: var(--color-grey-225-10-95);
   --header-icon-fill-color: var(--color-grey-225-10-15);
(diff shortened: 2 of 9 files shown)


pinussilvestrus:
Next up would be to integrate it into dmn-js as well + have it prepared for form-js (cf. https://github.com/bpmn-io/form-js/pull/254)

src/components/Placeholder.js
```diff
@@ -0,0 +1,24 @@
...
+        <p class="bio-properties-panel-placeholder-text">{ text }</p>
+      </section>
+    </div>
+  );
+}
+
+const noop = () => null;
+
+export const NOOP_PROVIDER = {
+  getEmptyText: noop,
```
nikku, line 20: Any reason we don simplify this to `placeholder = { text, icon? }` and make a provider offer `placeholderProvider.empty()` and `placeholderProvider.multiple()`. For me this looks like a duplication to have `get*Text` and `get*Icon` in the API.

nikku (changes requested):

src/components/Placeholder.js
pinussilvestrus: Good thing, let's try to simplify this 👍

pinussilvestrus: I updated it accordingly, the placeholder has now to provide `.empty()` and `.multiple()` 👍

nikku (approved):
Looks good to me.

pinussilvestrus:
Ah, sorry I missed the design review from @andreasgeier, let's maybe wait for the release until he gives his quick 👍

nikku:
> https://github.com/bpmn-io/properties-panel/pull/152#issuecomment-1132610043

I think it is fine if @andreasgeier follows up after the fact next week.

nikku:
(You have my poormans design review :wink:).

nikku:
One thing that confuses me (but it is bpmn-properties-panel wise) is the message for multiple elements.

Could we just simplify it to:

> Select a single element to edit its properties? 

Why would we want to hint users on `Context Pad`, something they may or may not know as a term?

christian-konrad:
@nikku because we now have the Multi-Element Context Pad and there is nothing to do for the user in the properties panel :)
The name "Context Pad" may be unknown, but then just let us introduce it. I think it's no regret.

nikku:
I disagree with https://github.com/bpmn-io/properties-panel/pull/152#issuecomment-1132634450. We should keep our messaging as simple as possible. Interaction with elements via the canvas is obvious for users; properties panel does not need to help out with that.

What is needed for users is to understand that they _need to select exactly a single element_ for properties editing to take place.

pinussilvestrus:
Thanks for your feedback @nikku. Let's move this over to the actual BPMN PR: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/687#issuecomment-1132684836

PR #154, opened 2022-06-01
chore(TextField): use controlled input

### LINKED ISSUE #146: Debounced Input Sometimes Loses Input

marstamm:
closes #146

test it with 
```
npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel -l bpmn-io/properties-panel#input-focus
```

### DIFF
--- a/src/components/entries/TextArea.js
+++ b/src/components/entries/TextArea.js
@@ -2,7 +2,9 @@ import Description from './Description';
 import FeelIcon from './FeelIcon';
 
 import {
-  useMemo
+  useEffect,
+  useMemo,
+  useState
 } from 'preact/hooks';
 
 import classnames from 'classnames';
@@ -21,9 +23,24 @@ function TextArea(props) {
     monospace
   } = props;
 
-  const handleInput = useMemo(() => {
+  const [localValue, setLocalValue] = useState(value);
+
+  const handleInputCallback = useMemo(() => {
     return debounce(({ target }) => onInput(target.value.length ? target.value : undefined));
   }, [ onInput, debounce ]);
+
+  const handleInput = e => {
+    handleInputCallback(e);
+    setLocalValue(e.target.value);
+  };
+
+  useEffect(() => {
+    if (value === localValue) {
+      return;
+    }
+
+    setLocalValue(value);
+  }, [value]);
 
   return (
     <div class="bio-properties-panel-textarea">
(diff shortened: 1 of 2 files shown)


marstamm:
As we just changed from uncontrolled to controlled input, I can't think of a test case that would behave differently. Let me know if you have any Idea on how to test for this weird glitch.

nikku:
I can still (rarely) reproduce it on `<textarea>`. I can no more reproduce it on text fields.

marstamm:
I was not aware that it was also am issue in Textareas and did not touch them. I will apply the same fix to Textareas :smile:

marstamm:
I could reproduce it on TextAreas and added the same pattern to it. Could not reproduce it after.

ca4f88661d48d4131552e0357e627f50cc28758c

If we find more places where this pattern is needed, it would make sense to extract it to a custom hook, but for just 2 Components I think it is okay to have the code duplicated.

pinussilvestrus:
> If we find more places where this pattern is needed, it would make sense to extract it to a custom hook, but for just 2 Components I think it is okay to have the code duplicated.

Could this also happen on [Number](https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/NumberField.js) (used in form-js) and [Simple](https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/Simple.js) entries (used in Camunda Platform 8 Inputs & Outputs - type = List)?

marstamm:
I think it could happen there as well (just looking at the code).

I would propose to merge this one as a fix to close the reported bug and create a follow-up ticket to migrate all components to controlled inputs.

nikku (approved):
Looks good. :+1:

nikku:
@marstamm Agreed. Please follow up (or create a follow up issue).

marstamm:
Follow-up: https://github.com/bpmn-io/properties-panel/issues/155

PR #159, opened 2022-07-01
Make all input fields controlled

### LINKED ISSUE #155: Make all inputs controlled
__What should we do?__

- All input fields should be controlled inputs
- Follow-up from https://github.com/bpmn-io/properties-panel/pull/154

__Why should we do it?__

https://github.com/bpmn-io/properties-panel/issues/146 was only reported on Text inputs so far, but is also possible for other input types. To prevent input loss, we should convert them all to the pattern introduced with https://github.com/bpmn-io/properties-panel/pull/154

smbea:
Closes #155

### DIFF
--- a/src/components/entries/Checkbox.js
+++ b/src/components/entries/Checkbox.js
@@ -3,6 +3,11 @@ import {
   useShowErrorEvent
 } from '../../hooks';
 
+import {
+  useEffect,
+  useState
+} from 'preact/hooks';
+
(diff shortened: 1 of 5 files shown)


nikku (approved):
I can verify this fixes the issue. 

Awesome!

nikku:
For good measure we could likely have extracted that into a shared hook. 

Let's get this merged first though, and consider to follow up.

PR #160, opened 2022-07-04
531 toggle linting spike

philippfromme:
This pull request introduces a way of setting errors through the `propertiesPanel.setError` event. The `propertiesPanel.showEntry` now expects the ID of an entry. The `propertiesPanel.showError` event was removed.

### Breaking Changes

* useShowEntryEvent hook expects ID instead of callback
* useShowErrorEvent hook and `propertiesPanel.showError` removed, add errors through context instead

---

Related to https://github.com/bpmn-io/internal-docs/issues/531


philippfromme:
> Just one question before merging this: we now kind of stick to the way we use IDs to reference fields. Maybe we stuck to it before, but now it feels that it will be harder to change this in the future, as linting is heavily relying on this.
> 
> Why I'm asking: We had some discussions in the past to move from IDs to something more robust (cf. [bpmn-io/bpmn-js-properties-panel#521](https://github.com/bpmn-io/bpmn-js-properties-panel/issues/521)). I assume we do not proceed on this for now?

Thanks for mentioning that. I'd say we can still refactor the way IDs are generated in the future. Of course, we'd have to adjust `@camunda/linting` in that case. In any case, we'll probably have to make `@bpmn-io/properties-panel` a peer dependency of `@camunda/linting`. I think that's fine.

pinussilvestrus (approved):

PR #173, opened 2022-08-03
fix(FEEL): show syntax errors

marstamm:
try it out using 

```
npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel -l bpmn-io/properties-panel#feel-editor-syntax-error
```

image

requires https://github.com/bpmn-io/feel-editor/pull/10

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -517,7 +517,12 @@ textarea.bio-properties-panel-input {
   background-color: var(--input-error-background-color);
 }
 
-.bio-properties-panel-entry.has-error .bio-properties-panel-input:focus {
+.bio-properties-panel-entry.has-error .bio-properties-panel-feel-indicator {
+  border-color: var(--input-error-border-color);
+}
+
+.bio-properties-panel-entry.has-error .bio-properties-panel-input:focus,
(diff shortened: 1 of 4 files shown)


philippfromme:
I'm reviewing this. 👀

philippfromme:
When there's an error, and I start typing, the error will only disappear after denounced linting has finished. I would expect the error to disappear when I start typing.

brave_C37xTP6SfT

On the other hand, in the Camunda Modeler, linting errors coming from the BPMN linter will behave the same. 🤔

marstamm:
I would keep it consistent with what is shown in the editor. As CodeMirror still shows error markers after we started typing, we should keep the error message in the properties panel as well.

VSCode seems to do the same, but with a much smaller delay. We could think about lowering the delay until linting is executed (default is 750ms), but that would be done on the `@bpmn-io/feel-editor` part

marstamm:
@philippfromme , should we merge and include this one in the next modeler release? The improvements mentioned in my previous comment have to be done as a follow-up in the `feel-editor` repo anyway.

pinussilvestrus (approved):
This looks good to me! Let's merge, any further improvements can be improved as follow-up (as @marstamm  suggested).

PR #175, opened 2022-08-16
fix(hooks): set sticky correctly on element change

marstamm:
related to https://github.com/bpmn-io/bpmn-js-properties-panel/issues/726
(see issue for why this was broken)

### DIFF
--- a/test/spec/hooks/useStickyIntersectionObserver.spec.js
+++ b/test/spec/hooks/useStickyIntersectionObserver.spec.js
@@ -11,11 +11,8 @@ describe('hooks/userStickyIntersectionObserver', function() {
     global.IntersectionObserver = OriginalIntersectionObserver;
   });
 
-  /**
-   * @pinussilvestrus Note: we exclude the sticky header feature until we
-   * find a proper fix for https://github.com/bpmn-io/bpmn-js-properties-panel/issues/726
-   */
-  it.skip('should observe', async function() {
+
+  it('should observe', async function() {
 
     // given
     const observeSpy = sinon.spy();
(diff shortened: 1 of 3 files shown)


test/spec/hooks/useStickyIntersectionObserver.spec.js
```diff
@@ -113,14 +140,30 @@ describe('hooks/userStickyIntersectionObserver', function() {
...
 
-  global.IntersectionObserver = class IntersectionObserver {
-    constructor() {}
+  const callbacks = [];
+
+  function triggerCallbacks(args) {
+    callbacks.forEach(callback => callback(args));
+  }
+
+  const MockObserver = class IntersectionObserver {
```
barmac, line 163: `extends` is necessary if you want the mock to have IntersectionObserver's APIs.

```suggestion
  const MockObserver = class extends IntersectionObserver {
```

marstamm: I don't think we want to extend it, rather replace it with our own implementation. But you are right that this looks wrong an could just be
```suggestion
  class MockObserver {
```

nikku:
Try out via 

```
npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel -l bpmn-io/properties-panel#fix-sticky-headers
```

marstamm:
Good catch @nikku. This is a z-index issue that we should fix in the headers. The FEEL marker already uses z-index 1, so the header should use something higher. I'll adjust the PR

nikku (approved):

PR #186, opened 2022-09-19
Do Not Account for Async Rendering of Properties Panel

philippfromme:
# Context

Due to the async rendering of React applications subscribing to events within a component isn't reliable as the event might already have been fired when the component is rendered and subscribes. As discussed with @nikku events should be subscribed to outside of the React application and relevant information should be passed to the application through props.

### Sketch

```javascript
function renderApp(props) {
  render(<App {...props} />, document.getElementById("root"));
}

eventBus.on('selection.changed', ({ newSelection }) => renderApp({ element: getElement(newSelection) });

eventBus.on('propertiesPanel.showEntry', ({ id }) => renderApp({ showEntry: id });
```

This requires a bigger implementation effort. In the meantime, to fix this issue, the following changes have been made:

* remove `useEventBuffer` hook
* use `useEffect` instead of `useLayoutEffect`

These changes remove some of the magic 🧙🏻‍♂️ that was added to work around the issue described above. The async rendering will be accounted for by https://github.com/camunda/linting/pull/19.

---

Related to https://github.com/camunda/camunda-modeler/issues/3144


nikku (approved):
Less :magic_wand:, yes!

PR #187, opened 2022-09-20
chore(CHANGELOG): update to v0.21.0

philippfromme:
I believe that this should be a breaking release even though the refiring of events was an internal the user wasn't necessarily aware of. 

# Changes

* `CHORE`: do not refire events during effect phase ([#186](https://github.com/bpmn-io/properties-panel/pull/186))

### BREAKING CHANGES

* events are not refired during effect phase anymore, you might have to use timeouts to ensure the events are being caught


barmac (approved):
Makes sense to me.

PR #191, opened 2022-10-21
feat: expose onFocus and onBlur functions on TextAreaEntry

zhGio:
The inner component `TextArea`  spreads the `props.onBlur` and `props.onFocus` functions to the native textarea component:
``` 
<textarea
        ....
        onFocus={ props.onFocus }
        onBlur={ props.onBlur }
        ... />
```
But the outer component `TextAreaEntry` never actually exposes the same props to the consumer.

### DIFF
--- a/src/components/entries/Checkbox.js
+++ b/src/components/entries/Checkbox.js
@@ -16,7 +16,9 @@ function Checkbox(props) {
     label,
     onChange,
     disabled,
-    value = false
+    value = false,
+    onFocus,
+    onBlur
   } = props;
 
   const [ localValue, setLocalValue ] = useState(value);
@@ -46,6 +48,8 @@ function Checkbox(props) {
         ref={ ref }
         id={ prefixId(id) }
         name={ id }
+        onFocus={ onFocus }
+        onBlur={ onBlur }
         type="checkbox"
         class="bio-properties-panel-input"
         onChange={ handleChange }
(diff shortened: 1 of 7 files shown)


marstamm:
Hi @zhGio , thank you for opening this PR. It seems that this is missing from all components except `simple`.

Do you think it makes sense to add them to https://github.com/bpmn-io/properties-panel/blob/6075388ca0f523548d95f659e02a27a83142f24a/src/components/entries/TextField.js#L143-L151 and https://github.com/bpmn-io/properties-panel/blob/6075388ca0f523548d95f659e02a27a83142f24a/src/components/entries/FEEL/Feel.js#L313-L325 as well?

zhGio:
Hi, @marstamm 

Yeah, I think it would make sense to expose those events to consumers on all components that render native components capable of focus/blur. I just ran into the issue of not having them on the `TextArea` while trying to build a feature.
Do you want me to add them to the rest of the components in this PR, open up a new PR, or is that something you guys would do? (if you agree those events should be exposed in the first case :))

marstamm:
Hi, 
Yes, I think these should be exposed. I see no reason why we would hide them when they can be useful :)
It would be a great help if you could add them to this PR, so we have consistent API across all fields. Thanks again for your contribution!

zhGio:
Hi @marstamm, updated this PR with a lot more changes so if you could please review :)

I've manually tested some of the examples and ran all the tests, seems ok to me.

marstamm:
Thank you for the follow-up. CI is happy as well, so let's merge it 🚀

Issue #169, opened 2022-07-20
Implement optgroup in SelectEntry


Ichigo85:
__Is your feature request related to a problem? Please describe__

I am developing (and/or rewriting) my own properties panel module, specifically for Camunda Modeler but not necessary related to it. While developing a select box I noted that's not possible to define a select box with the optgroup tag and it would be a bit hard to develop it on my own.

__Describe the solution you'd like__

I would like to add a prop to the SelectEntry component to allow optgroups while adding options

__Describe alternatives you've considered__

Another alternative was to create a "parent" select box for my groups, but since there is no model property for them because it's a way to set different delegateExpression values, a single select with optgroups would do the trick.

__Additional context__

Here is a screenshot of my selectbox in Camunda Modeler 4.
![image](https://user-images.githubusercontent.com/27808087/179940077-641d264f-5f2a-40fa-b221-4ee45758a8eb.png)

pinussilvestrus:
Makes sense, thanks for opening 👍

PR #204, opened 2022-12-24
feat: add optgroup option for select

etozhealkhipce:
This PR adds `optgroup` props to the `Select.js` component.

You will be able to use this by adding  `getGroups` function, which returns an array of objects with `key` and `label` fields in the select component props.

For proper functioning, you should add a `groupKey` field into the option(s) object, which, in its turn, should be equal to `key` field of group object

Example:

```
function ExampleComponent(props) {
  const { element } = props

  const getGroups = (element) => {
    return [
      {
        key: 'first',
        label: 'first test group'
      },
      {
        key: 'second',
        label: 'second test group'
      }
    ]
  }

  const getOptions = (element) => {
    return [
      {
        groupKey: 'first',
        label: 'Test option 1',
        value: 'test1'
      },
      {
        groupKey: 'first',
        label: 'Test option 2',
        value: 'test2',
        disabled: true
      },
      {
        groupKey: 'second',
        label: 'Test option 3',
        value: 'test3'
      },
      {
        label: 'Test option 4 without group',
        value: 'test4'
      }
    ]
  }

  return [
    SelectEntry({
      element,
      id: 'test',
      getGroups,
      label: translate('test'),
      getValue: () => null,
      setValue: (value) => console.log(value),
      getOptions
    })
  ]
}
```

image

Closes https://github.com/bpmn-io/properties-panel/issues/169

### DIFF
--- a/src/components/entries/Select.js
+++ b/src/components/entries/Select.js
@@ -13,7 +13,8 @@ import {
 import Description from './Description';
 
 /**
- * @typedef { { value: string, label: string, disabled: boolean } } Option
+ * @typedef { { value: string, label: string, disabled: boolean, groupKey: (string|undefined) } } Option
+ * @typedef { { key: string, label: string } } OptionGroup
  */
 
 /**
@@ -29,6 +30,7 @@ import Description from './Description';
  * @param {Array<Option>} [props.options]
  * @param {string} props.value
  * @param {boolean} [props.disabled]
+ * @param {Array<OptionGroup>} [props.groups]
  */
 function Select(props) {
   const {
@@ -36,6 +38,7 @@ function Select(props) {
     label,
     onChange,
     options = [],
+    groups = [],
     value = '',
     disabled,
     onFocus,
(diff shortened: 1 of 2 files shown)


barmac:
@etozhealkhipce We need you to sign the CLA if you want your contribution to be merged.

etozhealkhipce:
> 

Ok, how can I do that?

etozhealkhipce:
> @etozhealkhipce We need you to sign the CLA if you want your contribution to be merged.

Ok, I signed it.

barmac:
Cool, thanks.

I am not convinced about the proposed API. We require implementors to keep getGroups and getOptions in sync while the data structure could enforce it. Imagine if we keep `getOptions` as the only available method:

```javascript
const getOptions = (element) => {
  return [
    {
      label: 'first',
      children: [
        {
          label: 'Test option 1',
          value: 'test1'
        },
        {
          label: 'Test option 2',
          value: 'test2',
          disabled: true
        },
      ]
    },
    {
      label: 'second',
      children: [
        {
          label: 'Test option 3',
          value: 'test3'
        },
      ]
    },
    {
      label: 'Test option 4 without group',
      value: 'test4'
    }
  ]
}
```

So no need to care about `groupKey`, and as a bonus the ordering of options is explicit.

barmac (changes requested):
See the comment above.

etozhealkhipce:
> See the comment above.

I changed API according to your comment.

Please, check the changes.

barmac:
Great, thank you!

PR #213, opened 2023-01-17
feat: enable number field validation and linting errors

### LINKED ISSUE #206: Improve validation support for `NumberFieldEntry`

Currently, dealing with number fields can be a little bit confusing. While there is browser built-in validation for number fields, how it is displayed is browser dependent and at least in Firefox requires a hover on top of the field, which is pretty hard to find.

image

I don't think there exists a way to pipe the validation errors through to the UI via code (I could be wrong), though I know that via CSS we can at least apply some styling rules if the field is invalid, this might be useful. But in practice, I'd still rather have a similar validation experience as with the regular text fields. 

Allow the supplying of a validation function for the `NumberFieldEntry` component, mirroring how it is done for `TextFieldEntry`.

Rely solely on browser default validation, unfortunately, the way browsers do this lacks a bit of clarity

__Additional context__

Initial comment that shows the confusion people experience with this: https://github.com/bpmn-io/form-js/pull/490#pullrequestreview-1245538661

@nikku @philippfromme I'm happy to implement this myself, if you think this is a positive addition.

Skaiir:
Closes #206

image

### DIFF
--- a/src/components/entries/NumberField.js
+++ b/src/components/entries/NumberField.js
@@ -1,10 +1,19 @@
+import Description from './Description';
+
 import {
   useEffect,
   useMemo,
   useState
 } from 'preact/hooks';
 
-import Description from './Description';
+import classnames from 'classnames';
+
+import { isFunction } from 'min-dash';
+
+import {
+  useError,
+  usePrevious
+} from '../../hooks';
 
 function NumberField(props) {
 
@@ -88,6 +97,7 @@ function NumberField(props) {
  * @param {Function} props.onFocus
  * @param {Function} props.onBlur
  * @param {String} props.step
+ * @param {Function} props.validate
  */
 export default function NumberFieldEntry(props) {
   const {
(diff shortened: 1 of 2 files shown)


src/components/entries/NumberField.js
```diff
@@ -103,25 +113,66 @@ export default function NumberFieldEntry(props) {
...
     step,
     onFocus,
-    onBlur
+    onBlur,
+    validate
   } = props;
 
-  const value = getValue(element);
+  const [ cachedInvalidValue, setCachedInvalidValue ] = useState(null);
+  const globalError = useError(id);
```
pinussilvestrus, line 121: This also enables setting the errors from outside via linting, which is nice given we might want to enable it for quite some number properties in forms 👍

src/components/entries/NumberField.js
```diff
@@ -103,25 +113,66 @@ export default function NumberFieldEntry(props) {
...
+    let newValidationError = null;
+
+    if (isFunction(validate)) {
+      newValidationError = validate(newValue) || null;
+    }
+
+    if (newValidationError) {
+      setCachedInvalidValue(newValue);
+    } else {
+      setValue(newValue);
```
pinussilvestrus, line 146: Do we have a test for this? `it('should not set value on error')`.

Skaiir: We now do :)

test/spec/components/NumberField.spec.js
```diff
@@ -182,6 +184,120 @@ describe('<NumberField>', function() {
...
+  });
+
+
+  describe('validation', function() {
+
+    it('should set valid', function() {
+
+      // given
+      const validate = (v) => {
+        if (v === 'bar') {
```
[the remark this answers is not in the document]
Skaiir: I copied the tests over and forgot to clean this up because the test wasn't failing. I've changed it to be in line with the previous.

pinussilvestrus (approved):
LGTM 👍

pinussilvestrus:
@bpmn-io/modeling-dev feel free to still provide your feedback!

nikku (approved):
Effectively what we're doing here is enforcing _strict_ validation. An invalid field never ends up in the model.

I can see [`TextField`](https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/TextField.js) doing the same thing.

Consistency wise this looks like a good contribution.

PR #214, opened 2023-02-02
feat(textarea): provide auto-resizeable textarea

smbea:
Related to https://github.com/bpmn-io/bpmn-js-properties-panel/issues/705, specifically https://github.com/bpmn-io/bpmn-js-properties-panel/pull/864#discussion_r1094405640

capture o3BxBW_optimized

https://github.com/bpmn-io/bpmn-js-properties-panel/pull/864 pins this version of `properties-panel` and allows to test end-to-end

### DIFF
--- a/test/spec/components/TextArea.spec.js
+++ b/test/spec/components/TextArea.spec.js
@@ -305,7 +305,34 @@ describe('<TextArea>', function() {
 
   describe('auto resize', function() {
 
-    it('should resize textarea', function() {
+    it('should resize initially', function() {
+
+      // given
+      const result = createTextArea({
+        container,
+        id: 'textarea',
+        getValue() {
+          return `
+HALLO
+WELT
+WIE
+GEHTS
+`;
+        },
+        autoResize: true
+      });
+
+      const input = domQuery('.bio-properties-panel-input', result.container);
+
+      // when
+      const initialHeight = input.clientHeight;
+
+      // then
+      expect(initialHeight).to.be.greaterThan(60);
+    });
+
+
+    it('should resize on input', function() {
 
       // given
       const result = createTextArea({
--- a/src/components/entries/TextArea.js
+++ b/src/components/entries/TextArea.js
@@ -2,6 +2,7 @@ import Description from './Description';
 
 import {
   useEffect,
+  useLayoutEffect,
   useMemo,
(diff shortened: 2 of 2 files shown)


nikku (changes requested):
Great start I think. 

Usability wise I'm not sold on the `max-height` usage and the behavior it triggers: As a user I want to resize the panel to the contents. `max-height` prevents me to manually resize:

capture l4evWU_optimized

Instead I'd ensure that we assert the maximum height until which we _autoresize_ via JavaScript, not CSS. :arrow_right: Users can still resize on top.

I'd like to make sure that the scroll bar, if possible is hidden. This can be accomplished by making the panel 2px larger than the scroll height, I believe.

Both can be seen in action here:

capture WZnbpf_optimized

Last I'd assume that a reasonable default high is assumed on initial render. If I display multi-line text then we should apply the same auto-resize rules on initial rendering, or the result is unexpected:

capture 5Bcdfh_optimized

nikku:
Pushed https://github.com/bpmn-io/properties-panel/pull/214/commits/5ec10920741ed6ca85420b2b9128998fd479ac8f to address the first two issues mentioned [here](https://github.com/bpmn-io/properties-panel/pull/214#pullrequestreview-1281474171). 

Remaining: Auto-resize on initial render.

nikku:
Fixed via https://github.com/bpmn-io/properties-panel/pull/214/commits/019bd8fb33ee00049bb4939fed1ae973082879c4. 

capture Djge9E_optimized

nikku (commented):
Good stuff :clap:

nikku (approved):

test/spec/components/TextArea.spec.js
```diff
@@ -303,6 +303,65 @@ describe('<TextArea>', function() {
...
+      // given
+      const result = createTextArea({
+        container,
+        id: 'textarea',
+        getValue() {
+          return `
+HALLO
+WELT
+WIE
+GEHTS
```
marstamm, line 319: 👋 🌏

philippfromme: Geht so.

nikku: :smile:

PR #216, opened 2023-02-07
fix: do not grow autoresized text area on first input

barmac:

### DIFF
--- a/src/components/entries/TextArea.js
+++ b/src/components/entries/TextArea.js
@@ -20,7 +20,10 @@ function resizeToContents(element) {
   }
 
   element.style.height = 'auto';
-  element.style.height = Math.min(element.scrollHeight, 150) + 'px';
+
+  // a 2px pixel offset is required to prevent scrollbar from
+  // appearing on OS with a full length scroll bar (Windows/Linux)
+  element.style.height = `${ element.scrollHeight + 2 }px`;
 }
 
 function TextArea(props) {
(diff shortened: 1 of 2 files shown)


nikku:
Thanks for looking into this. I don't understand why the `scrollHeight` check should be necessary. We enforce a minimum height via CSS. Removing it (https://github.com/bpmn-io/properties-panel/pull/216/commits/b1b8a108fe9299171702429952a26fc2e354a31d) the test still passes.

There is a couple of nuances that we need to be aware on non MacOS systems regarding scrollbars. Scroll height will always be < then offsetHeight once resized, hence we must make the actual height magic `2px` bigger than scroll height:

capture QmX6JA_optimized

If we don't do that then a pesky scrollbar appears every time we auto-resize.

Finally, revisiting this I find that our cap of `150px` is an arbitrary one, and dangerous, too. We'd need to properly handle manual user resize in order to turn that into a meaningful thing. I've removed it via https://github.com/bpmn-io/properties-panel/pull/216/commits/29e9612879ca7e480f2dfa001751ab72fc862777.

Please check if `rows=1` was the actual fix for the issue. Consider squashing my PRs then.

nikku (changes requested):
I think this can be simplified, and `+ 2px` must be kept: https://github.com/bpmn-io/properties-panel/pull/216#issuecomment-1421417303.

barmac:
It still grows on MacOS / Chrome. It looks like the +2px causes this:

https://user-images.githubusercontent.com/28307541/217468875-a9b90fd8-f7c3-477d-94a7-e395008fa71f.mov

barmac:
> Please check if rows=1 was the actual fix for the issue. Consider squashing my PRs then.

`rows=2` makes it grow to 2 lines height on my machine.

PR #224, opened 2023-02-27
fix: When running in shadowRoot, after debounce keep event target.

### LINKED ISSUE #222: When running in shadowDom, debounce causes evnet objects to reset target property.
### Describe the Bug

I use bpmn-js in shadowDom and [bpmn-js-properties-panel](https://kgithub.com/bpmn-io/bpmn-js-properties-panel). 
There is an error in the console when I try to modify properties in the property-panel.

image
image

After testing, I found that the reason was that setTmeout in debounce caused the event.target to be reset.
[event-target-changed-in-settimeout-in-shadow-dom](https://stackoverflow.com/questions/63607966/event-target-changed-in-settimeout-in-shadow-dom).

### Steps to Reproduce

https://codesandbox.io/s/distracted-leavitt-n8gxfn?file=/src/index.js

### Environment

- Host: Chrome 110.0.5481.100
- OS: Windows 10
- Library version: [e.g. 2.0.0]
  - bpmn-js-properties-panel: 1.17.2
  - @bpmn-io/properties-panel: 1.3.1

siaikin:
Closes #222

### DIFF
--- a/src/features/debounce-input/debounceInput.js
+++ b/src/features/debounce-input/debounceInput.js
@@ -2,6 +2,7 @@ import {
   debounce,
   isNumber
 } from 'min-dash';
+import { keepEventTarget } from '../../utils/shadow-dom';
 
 const DEFAULT_DEBOUNCE_TIME = 300;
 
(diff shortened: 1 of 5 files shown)


siaikin:
@nikku any help about this fail https://github.com/bpmn-io/properties-panel/actions/runs/4280345868/jobs/7611120251#step:7:1, 哦i have no idea.

PR #225, opened 2023-03-03
feat: use global layout context as single source of truth

marstamm:
- Groups: don't use local state in [`useLayoutState`](https://github.com/bpmn-io/properties-panel/blob/e01dd051aba1b4998bcc6f7a591ad7cdf8fc4cf7/src/hooks/useLayoutState.js#L26). Instead, always use the Context value, so outside changes in the layout are reflected in the Groups.
- PropertiesPanel (root): React to `layoutConfig` changes, e.g. from other tabs. Currently, only the initial value is used.

related to https://github.com/camunda/camunda-modeler/issues/2638


philippfromme (approved):
✅

PR #226, opened 2023-03-10
feat: feelers editor

Skaiir:
Related to [team-hto/141](https://github.com/camunda/team-hto/issues/141)

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -446,6 +446,21 @@
   text-decoration: underline;
 }
 
+.bio-properties-panel-feelers-editor.bio-properties-panel-input {
+  padding: 0;
+}
+
+.bio-properties-panel-feelers-input .cm-editor
+{
+  min-height: 32px;
+  max-height: 215px;
+  background-color: transparent;
+}
+
+.bio-properties-panel-feelers-input .cm-editor.cm-focused {
+  outline: none;
+}
+
 .bio-properties-panel-input {
   padding: 3px 6px 2px;
   border: 1px solid var(--input-border-color);
--- a/src/components/entries/templating/FeelTemplating.js
+++ b/src/components/entries/templating/FeelTemplating.js
@@ -18,7 +18,9 @@ const noop = () => { };
  * @param {String} props.label
  * @param {Function} props.getValue
  * @param {Function} props.setValue
+ * @param {Function} props.tooltipContainer
  * @param {Function} props.validate
+ * @param {Function} props.show
  */
 export default function FeelTemplatingEntry(props) {
   const {
--- a/src/components/entries/templating/FeelTemplating.js
+++ b/src/components/entries/templating/FeelTemplating.js
@@ -1,0 +1,206 @@
+import { useCallback, useEffect, useState, useMemo, useRef } from 'preact/hooks';
+import { usePrevious, useStaticCallback, useShowEntryEvent } from '../../../hooks';
+import { isFunction } from 'min-dash';
+import { useError } from '../../../hooks';
+import classnames from 'classnames';
+import Description from '../Description';
+import FeelTemplatingEditor from './FeelTemplatingEditor';
+
+const noop = () => { };
+
+/**
+ * @param {Object} props
+ * @param {Object} props.element
+ * @param {String} props.id
+ * @param {String} props.description
+ * @param {Boolean} props.debounce
+ * @param {Boolean} props.disabled
+ * @param {String} props.label
+ * @param {Function} props.getValue
+ * @param {Function} props.setValue
+ * @param {Function} props.validate
+ */
+export default function FeelTemplatingEntry(props) {
+  const {
+    element,
+    id,
+    description,
+    debounce,
+    disabled,
+    label,
+    getValue,
+    setValue,
+    tooltipContainer,
+    validate,
+    show = noop,
+  } = props;
+
+  const [ cachedInvalidValue, setCachedInvalidValue ] = useState(null);
+  const [ validationError, setValidationError ] = useState(null);
+  const [ localError, setLocalError ] = useState(null);
+
+  let value = getValue(element);
+
+  const previousValue = usePrevious(value);
+
+  useEffect(() => {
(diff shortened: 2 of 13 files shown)


Skaiir:
I am struggling to get these tests working. There's some interference between tests and I cannot find the root cause except maybe that it's related to asynchronous programming. 

Anyways, putting this up for review already because otherwise, it'll take aeons.

test/spec/components/FeelTemplating.spec.js
```diff
@@ -0,0 +1,608 @@
...
+
+      // given
+      const updateSpy = sinon.spy();
+
+      const result = createTemplatingEntry({ container, setValue: updateSpy });
+
+      const input = domQuery('[role="textbox"]', result.container);
+
+      // when
+      await act(() => input.textContent = 'Template {{x}}');
```
pinussilvestrus, line 71: Maybe this is not enough to make a proper update? In bpmn-js-properties-panel and form-js we use a helper to do the updates

```js
async function setEditorValue(editor, value) {
  await act(() => {
    editor.textContent = value;
  });

  // Requires 2 ticks to propagate the change
  await act(() => {});
}
```

Not sure whether this is causing the test errors though

pinussilvestrus: I pushed a fix. Probs to @marstamm finding this via https://github.com/bpmn-io/properties-panel/pull/225/commits/a4706eff443af23972719a10d75a35e6f99b0aff

pinussilvestrus:
@Skaiir I fixed the skipped tests and opened an issue for the feel editor to be able to reference a label: https://github.com/bpmn-io/feel-editor/issues/36

I'll have a review later today 👍

assets/properties-panel.css
```diff
@@ -446,6 +446,21 @@
   text-decoration: underline;
 }
 
+.bio-properties-panel-feelers-editor.bio-properties-panel-input {
+  padding: 0;
+}
+
+.bio-properties-panel-feelers-input .cm-editor
+{
```
pinussilvestrus, line 454:
```suggestion
.bio-properties-panel-feelers-input .cm-editor {
```

assets/properties-panel.css
```diff
@@ -446,6 +446,21 @@
...
+  min-height: 32px;
+  max-height: 215px;
```
pinussilvestrus, line 456: Why do we need restrictions here and not for the "normal" FEEL editor?

test/TestHelper.js
```diff
@@ -67,5 +67,10 @@ export async function expectNoViolations(node, options = {}) {
   });
 
   expect(results.passes).to.be.not.empty;
+
+  if (results.violations.length) {
+    console.log(JSON.stringify(results.violations, null, 2));
+  }
```
pinussilvestrus, line 73: 👍

pinussilvestrus: Why the stringification? Wouldn't it be enough to inspect it in the browser dev tools?

src/components/entries/templating/FeelTemplating.js
```diff
@@ -0,0 +1,206 @@
...
+ * @param {String} props.id
+ * @param {String} props.description
+ * @param {Boolean} props.debounce
+ * @param {Boolean} props.disabled
+ * @param {String} props.label
+ * @param {Function} props.getValue
+ * @param {Function} props.setValue
+ * @param {Function} props.validate
+ */
+export default function FeelTemplatingEntry(props) {
```
pinussilvestrus, line 23: Just to clarify: we do not bring the same optional vs. required behavior as we have in `FEEL` with the new templating, right? Once you use a FEEL templating entry, you can't turn it off.

src/components/entries/templating/FeelTemplating.js
```diff
@@ -0,0 +1,206 @@
...
+  const {
+    element,
+    id,
+    description,
+    debounce,
+    disabled,
+    label,
+    getValue,
+    setValue,
+    tooltipContainer,
```
pinussilvestrus, line 33: This is not added to the JSdocs. It would be interesting to see this live in action; I'd expect it's just a DOM element?

pinussilvestrus (commented):
Looks good from implementation perspective. Is there any way to properly test this out in action (form-js branch)?

src/components/entries/templating/FeelTemplating.js
Skaiir: In this first iteration anyways, yeah. But there are plans for this to change relatively soon ([link](https://github.com/camunda/product-hub/issues/163)).

src/components/entries/templating/FeelTemplating.js
Skaiir: Yeah so, I removed the JSdocs by accident from the codemirror editor. It's a DOM element or a DOM query that defines where tooltips are rendered, as we may want to render them at an entirely different layer than the editor.

test/TestHelper.js
Skaiir: Well, this would be visible in CI for example, and also for running all tests I find the console is a bit easier as it doesn't flood with text unlike karma in the browser.

assets/properties-panel.css
Skaiir: It's a design decision that was taken during a little UX session. In any case since we might merge the two editors, I think let's not worry about how the other one displays for now.

src/components/entries/templating/FeelTemplating.js
pinussilvestrus: Should we add it to the JSdocs of `FeelTemplatingEntry` then? Just to be consistent.

pinussilvestrus (commented):
✅ from my side 👍 I'll leave the chance for @bpmn-io/modeling-dev to also have a look.

nikku:
@marstamm and others this is where I think we'd benefit from [pluggable language editor support](https://github.com/bpmn-io/bpmn-js-properties-panel/issues/871).

Happy to incorporate it like this for the time being.

PR #231, opened 2023-03-22
fix(ListGroup): don't open on external changes

marstamm:
Found while working on https://github.com/bpmn-io/bpmn-js-example-data-extension/issues/1, cf. https://camunda.slack.com/archives/GP70M0J6M/p1679407286325119

This ensures that we only open the section when the change originated from the `add` button click:

Recording 2023-03-22 at 14 01 11

### DIFF
--- a/src/components/ListGroup.js
+++ b/src/components/ListGroup.js
@@ -247,8 +247,9 @@ export default function ListGroup(props) {
 
             const { id } = item;
 
-            // if item was added, open first or last item based on ordering
-            const autoOpen = newItemAdded && (shouldSort ? index === 0 : index === ordering.length - 1);
+            // if item was added, open it
+            // Existing items will not be affected as autoOpen is only applied on first render
+            const autoOpen = newItemAdded;
 
             return (
               <ListItem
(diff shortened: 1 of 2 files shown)


src/components/ListGroup.js
```diff
@@ -94,14 +101,14 @@ export default function ListGroup(props) {
 
       let newOrdering = ordering;
 
-      // open if not open and configured
-      if (!open && shouldOpen) {
+      // open if not open, configured and triggered by add button
+      if (addTriggered && !open && shouldOpen) {
```
nikku, line 105: Feels really like an awkward hack. Let's document it like that and add a `// TODO(marstamm): ...`.

marstamm: I added a `todo` and a bit of documentation to move the logic out of the ListGroup

nikku: Thanks @marstamm.

With these comments (and this commit as a reference) the changes look manageable. 

Let's try to keep our hacks per LOC as low as possible.

[image](https://xkcd.com/1513/)

PR #238, opened 2023-04-19
Fix flickering on desktop modeler tab change

marstamm:
Recording 2023-04-19 at 11 51 25

related to https://github.com/camunda/camunda-modeler/pull/3556

PR #241, opened 2023-05-04
fix(hooks): don't break on missing scrollContainer

marstamm:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to [`@bpmn-io/properties-panel`](https://github.com/bpmn-io/
 
 ___Note:__ Yet to be released changes appear here._
 
+## 1.8.3
+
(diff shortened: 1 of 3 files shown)


philippfromme:
So I assume you have the context and this is a regression caused by https://github.com/bpmn-io/properties-panel/commit/0ffe26bd7b11da61c3e3d7cd2abd651c7baa6b13? I'm asking because I didn't get this error.

marstamm:
Yes, this is a regression caused by #238 . Basically, we assume that the `scrollContainer` DOM query will return something, which is not the case when it is unmounted.

Here, I just added a check to verify that `scrollContainer` exists before we access it

philippfromme (approved):
✅

barmac:
Unfortunately, the test case is green even when the fix is removed.

PR #243, opened 2023-05-09
fix(Groups): update sticky observer on attach/detach

marstamm:
related to https://github.com/camunda/web-modeler/issues/4650

This PR:
 - subscribes to attach/detach events
 - then updates the scroll container
 - Removes/Adds an Intersection Observer to the new scrollContainer

### DIFF
--- a/src/hooks/useStickyIntersectionObserver.js
+++ b/src/hooks/useStickyIntersectionObserver.js
@@ -1,10 +1,14 @@
 import {
-  useEffect
+  useCallback,
+  useEffect,
+  useState
 } from 'preact/hooks';
 
 import {
   query as domQuery
 } from 'min-dom';
+
+import { useEvent } from './useEvent';
 
 
 /**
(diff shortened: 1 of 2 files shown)


src/hooks/useStickyIntersectionObserver.js
```diff
@@ -23,55 +27,64 @@ import {
...
+      return;
+    }
 
-    if (ref.current) {
-      const scrollContainer = domQuery(scrollContainerSelector);
+    const observer = new Observer((entries) => {
 
-      // TODO(@barmac): test this
-      if (!scrollContainer) {
+      // The ScrollContainer is unmounted, do not update sticky state
```
nikku, line 63:
```suggestion
      // scroll container is unmounted, do not update sticky state
```

src/hooks/useStickyIntersectionObserver.js
```diff
@@ -23,55 +27,64 @@ import {
...
+  useEffect(() => {
     const Observer = IntersectionObserver;
 
     // return early if IntersectionObserver is not available
     if (!Observer) {
       return;
     }
 
-    let observer;
+    // TODO(@barmac): test this
```
nikku, line 56: Sneaking in TODO for someone else? :see_no_evil:

nikku (commented):
Looks OK.

I like how the complexity is hidden, inside that hook.

PR #248, opened 2023-05-25
feat(feel): add number feel entry

pinussilvestrus:
Related to bpmn-io/form-js#653

Adds an optional FEEL number field entry. This enabled us to create number properties with FEEL support (e.g. `min`, `max`, `minLength`, `maxLength` in form-js, https://github.com/bpmn-io/form-js/pull/668).

### DIFF
--- a/src/components/entries/FEEL/Feel.js
+++ b/src/components/entries/FEEL/Feel.js
@@ -28,6 +28,8 @@ import FeelIcon from './FeelIcon';
 
 import { ToggleSwitch } from '../ToggleSwitch';
 
+import { NumberField } from '../NumberField';
+
 const noop = () => { };
 
 function FeelTextfield(props) {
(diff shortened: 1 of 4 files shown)


nikku:
Thanks @pinussilvestrus. I looked into the demo you provided and found one glitch: Entering a `.` breaks the editor focus (moves the cursor):

capture CFvskP_optimized

Expected behavior:

* character is disallowed _or_ character is passed, but an error message (invalid format) is shown

pinussilvestrus:
@nikku I decided to simply re-using the existing `NumberField` so we have a consistent behavior between FEEL and non-FEEL number entries. I'm not sure what's the expected behavior, but it should work the same as before now (without jumping cursor).

nikku (approved):
Works for me now.

PR #252, opened 2023-06-14
feat: propagate invalid values to `setValue`

marstamm:
Whether invalid values are persisted on errors or not should be a concern of the application. This PR introduces that we also propagate invalid values to the app.

Integrates into bpmn-js-properties-panel: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/938

related to https://github.com/bpmn-io/bpmn-js-properties-panel/issues/928
related to https://github.com/camunda/camunda-modeler/issues/3357

### DIFF
--- a/test/spec/components/Feel.spec.js
+++ b/test/spec/components/Feel.spec.js
@@ -277,6 +277,29 @@ describe('<FeelField>', function() {
         expect(error.innerText).to.eql('error');
       });
 
+
+      it('should pass error to `setValue`', function() {
+
+        // given
+        const setValueSpy = sinon.spy();
+        const validate = (v) => {
+          if (v === 'bar') {
+            return 'error';
+          }
+        };
+
+        const result = createFeelField({ container, validate, setValue: setValueSpy });
+
+        const entry = domQuery('.bio-properties-panel-entry', result.container);
+        const input = domQuery('.bio-properties-panel-input', entry);
+
+        // when
+        changeInput(input, 'bar');
+
+        // then
+        expect(setValueSpy).to.have.been.calledWith('bar', 'error');
+      });
+
     });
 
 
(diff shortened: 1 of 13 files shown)


philippfromme (commented):
So we now validate twice? We could add an optional `error` parameter to `setValue` instead.

marstamm:
> So we now validate twice?

Not necessarily. In most cases, the validation is only done as linting and not performed on `setValue`. Only for IDs we will perform the check twice.

I like the Idea to also provide the validation error in the `setValue` callback so we don't need to validate twice. I'll add it

marstamm:
Added the suggested improvement via [61a1dbb](https://github.com/bpmn-io/properties-panel/pull/252/commits/61a1dbb1a8699b684bd164e1dcdaf79274c1a8ce)

test/spec/components/Feel.spec.js
```diff
@@ -277,6 +277,29 @@ describe('<FeelField>', function() {
         expect(error.innerText).to.eql('error');
       });
 
+
+      it('should pass error to `setValue`', function() {
+
+        // given
+        const setValueSpy = sinon.spy();
+        const validate = (v) => {
```
philippfromme, line 285: I know we use this pattern a lot, but these validation functions are pointless. We know exactly if the validation is to return an error or not. So why not just do `const validate = () => 'error'`? I'll file a follow-up pull request simplifying this. Our tests should stay focused.

philippfromme (approved):
✅

test/spec/components/Feel.spec.js
marstamm: This might have been deliberate in the beginning (e.g. only set "invalid" when the check actually fails. However, we never do the `// assume - valid at the start` part, so we should be good to simplify

PR #256, opened 2023-07-12
feat: add error markers to groups

### LINKED ISSUE #180: Indicate errors also in the section headers

With our approach to end-to-end validation, we try to hint users to problems in the model in any potential place, so the user won't miss any problem. We list all problems in the problems panel and highlight all fields in the properties panel with erroneous content. However, if you collapse a section in the properties panel, you won't see that there is a problem in any field inside this section. Additionally, problems related to whole sections (such as using extension properties prior to 8.1) can not be displayed properly, resulting in redundant highlighting on a per-item basis (see https://github.com/camunda/linting/pull/14). 

In case there is any problem with any field inside a properties panel section, the "change" (respectively count) indicator should turn into an error sign, indicating that there is a problem to be fixed. In case the problem is with the whole section (such as with the extension properties example), an optional error message should be shown in the section header.

Bildschirmfoto 2022-09-05 um 20 52 01

Bildschirmfoto 2022-09-05 um 20 53 33

Bildschirmfoto 2022-09-05 um 21 06 07

__Additional context__

marstamm:
closes #180

Recording 2023-07-12 at 11 30 33

Unfortunately, `@bpmn-io/sr` does not work for this one. To test it out, link the properties-panel from `@camunda/linting` and change the import path for the properties-panel css to `@bpmn-io/properties-panel`: https://github.com/camunda/linting/blob/549ec9cfbed996845312015a804090ba08aeffb6/test/spec/modeler/Linting.spec.js#L40

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -62,6 +62,7 @@
   --arrow-hover-background-color: var(--color-grey-225-10-95);
 
   --dot-color: var(--color-grey-225-10-35);
+  --dot-color-error: var(--color-red-360-100-45);
 
   --list-badge-color: var(--color-white);
   --list-badge-background-color: var(--color-grey-225-10-35);
@@ -355,6 +356,10 @@
   background-color: var(--dot-color);
 }
 
+.bio-properties-panel-dot.error {
+  background-color: var(--dot-color-error);
+}
+
 /**
  * Lists
  */
@@ -380,6 +385,10 @@
   padding: 0 5px;
   margin: 5px;
   background-color: var(--list-badge-background-color);
+}
+
+.bio-properties-panel-list-badge.error {
+  background-color: var(--dot-color-error);
 }
 
 /**
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -356,7 +356,7 @@
   background-color: var(--dot-color);
 }
 
-.bio-properties-panel-dot.error {
+.bio-properties-panel-dot--error {
   background-color: var(--dot-color-error);
 }
 
@@ -387,7 +387,7 @@
   background-color: var(--list-badge-background-color);
 }
 
-.bio-properties-panel-list-badge.error {
+.bio-properties-panel-list-badge--error {
   background-color: var(--dot-color-error);
 }
 
--- a/src/components/Group.js
+++ b/src/components/Group.js
@@ -17,6 +17,7 @@ import {
 } from 'min-dash';
 
 import {
+  useErrors,
   useLayoutState
 } from '../hooks';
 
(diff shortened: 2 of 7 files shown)


assets/properties-panel.css
```diff
@@ -355,6 +356,10 @@
   background-color: var(--dot-color);
 }
 
+.bio-properties-panel-dot.error {
```
barmac, line 359: `.error` sounds super generic, and I am afraid that this can cause clashes in embedded scenarios. Imagine somebody uses the class name in their project's stylesheet (e.g. `.error { background: red }`). Then, it interferes with the properties panel unintentionally.
I'd rather try to use specific class names. This can be achieved via BEM (https://getbem.com/) which we sometimes use in our projects. So in this case we'd use:

```suggestion
.bio-properties-panel-dot--error {
```

src/components/Group.js
```diff
@@ -130,8 +139,22 @@ export default function Group(props) {
...
-  );
+function DataMarker(props) {
+  const {
+    edited,
+    hasErrors
+  } = props;
+
+  if (hasErrors) {
+    return (
+      <div title="Section contains an error" class="bio-properties-panel-dot error"></div>
```
barmac, line 150: Consequence of https://github.com/bpmn-io/properties-panel/pull/256/files#r1265083126

```suggestion
      <div title="Section contains an error" class="bio-properties-panel-dot bio-properties-panel-dot--error"></div>
```

src/components/Group.js
```diff
@@ -88,14 +93,18 @@ export default function Group(props) {
       'bio-properties-panel-group-header',
       edited ? '' : 'empty',
       open ? 'open' : '',
-      (sticky && open) ? 'sticky' : ''
+      (sticky && open) ? 'sticky' : '',
+      hasErrors ? 'error' : ''
```
barmac, line 97: Do we ever use this class?

assets/properties-panel.css
```diff
@@ -382,6 +387,10 @@
   background-color: var(--list-badge-background-color);
 }
 
+.bio-properties-panel-list-badge.error {
```
barmac, line 390: Same as https://github.com/bpmn-io/properties-panel/pull/256/files#r1265083126

barmac:
I've left a couple of comments regarding CSS. Once we fix them, I think we are on a good track to merge this.

marstamm:
Thank you for the review, I think the point is valid - we should not use non-namespaced classes. I implemented your suggestions with [4f17111](https://github.com/bpmn-io/properties-panel/pull/256/commits/4f1711140e59855e082a94bdb3fd6fc1eceef132)

src/components/Group.js
marstamm: Nope, I removed it

barmac (approved):
Well done!

assets/properties-panel.css
```diff
@@ -355,6 +356,10 @@
   background-color: var(--dot-color);
 }
 
+.bio-properties-panel-dot--error {
+  background-color: var(--dot-color-error);
```
barmac, line 360: BTW I just thought that if we wanted to make it even more resilient, we could re-assign the `dot-color` variable instead of changing an implementation detail of `background-color`:

```suggestion
  --dot-color: var(--dot-color-error);
```

Still, it's good enough as is.

PR #258, opened 2023-07-13
Support tooltips

### LINKED ISSUE #202: Support tooltips for entries

The properties panel currently supports descriptions to provide more help and context to a specific entry.

Example: BPMN properties panel

image

Example: Forms properties panel

image

If you want to provide more context without cluttering the UI or using links to refer to external documentation, we'd need something else to display bigger chunks of explanations and help texts.

Introduce the possibility of showing tooltips on specific entries, similar to descriptions, but only visible on hover.

image

Use native tooltips, which come with a lack of discoverability (cf. https://github.com/bpmn-io/bpmn-js-properties-panel/issues/725).

__Additional context__

Related to https://github.com/bpmn-io/bpmn-js-properties-panel/issues/725
Related to https://github.com/bpmn-io/form-js/issues/305

smbea:
Closes #202

* Adds tooltip components
* Adds it to input fields: Checkbox, Feel, Textfield, TextArea, ToggleSwitch

Can be tested against https://github.com/bpmn-io/bpmn-js-properties-panel/tree/tooltips

### DIFF
--- a/assets/properties-panel.css
+++ b/assets/properties-panel.css
@@ -11,12 +11,14 @@
   --color-grey-225-10-90: hsl(225, 10%, 90%);
   --color-grey-225-10-95: hsl(225, 10%, 95%);
   --color-grey-225-10-97: hsl(225, 10%, 97%);
+  --color-grey-0-0-22: hsl(0, 0%, 22%);
 
   --color-blue-205-100-35: hsl(205, 100%, 35%);
   --color-blue-205-100-45: hsl(205, 100%, 45%);
   --color-blue-205-100-50: hsl(205, 100%, 50%);
   --color-blue-205-100-95: hsl(205, 100%, 95%);
-
+  --color-blue-219-99-53: hsl(219, 99%, 53%);
+  --color-blue-218-100-74: hsl(218, 100%, 74%);
   --color-green-150-86-44: hsl(150, 86%, 44%);
 
   --color-red-360-100-40: hsl(360, 100%, 40%);
@@ -110,6 +112,12 @@
 
   --feel-indicator-background-color: var(--color-grey-225-10-90);
 
+  --tooltip-underline-color: var(--color-blue-219-99-53);
+  --tooltip-background-color: var(--color-grey-0-0-22);
+  --tooltip-link: var(--color-blue-218-100-74);
+  --tooltip-code-background-color: var(--color-grey-225-10-97);
+  --tooltip-code-border-color: var(--color-grey-225-10-85);
+  
   --text-size-base: 14px;
   --text-size-small: 13px;
   --text-size-smallest: 12px;
@@ -1121,4 +1129,53 @@ textarea.bio-properties-panel-input {
 .bio-properties-panel-feel-checkbox .bio-properties-panel-feel-entry:not(.feel-active) .bio-properties-panel-feel-container,
 .bio-properties-panel-feel-toggle-switch .bio-properties-panel-feel-entry:not(.feel-active) .bio-properties-panel-feel-container {
   margin-left: auto;
+}
+
+.bio-properties-panel-tooltip-wrapper {
+  text-decoration: underline;
+  text-decoration-style: dotted;
+  text-underline-offset: 2px;
+}
+
+.bio-properties-panel-tooltip {
+  display: flex;
+  color: var(--color-white, white);
+  position: fixed;
+  z-index: 1000;
+  padding-right: 6px;
+  max-width: 300px;
+  font-size: var(--text-size-small);
+  font-family: var(--font-family);
+}
+
+.bio-properties-panel-tooltip-content {
+  background-color: var(--tooltip-background-color);
+  padding: 16px;
+  border-radius: 2px;
+  font-weight: 400;
+}
+
+.bio-properties-panel-tooltip-content code,
+.bio-properties-panel-tooltip-content pre {
+  color: var(--description-color);
+  font-family: var(--font-family);
+  font-size: var(--text-size-small);
+  line-height: var(--line-height-condensed);
+  padding: 0 2px;
+  background-color: var(--tooltip-code-background-color);
+  border: 1px solid var(--tooltip-code-border-color);
+  border-radius: 3px;
+}
+
+.bio-properties-panel-tooltip-content a {
+  color: var(--tooltip-link);
+}
+
+.bio-properties-panel-tooltip .bio-properties-panel-tooltip-arrow {
+  width: 0; 
+  height: 0; 
+  border-top: 5px solid transparent;
+  border-bottom: 5px solid transparent;
+  border-left: 5px solid var(--tooltip-background-color);
+  margin-top: 16px;
 }
--- a/src/components/entries/Checkbox.js
+++ b/src/components/entries/Checkbox.js
@@ -9,6 +9,7 @@ import {
 } from 'preact/hooks';
 
 import Description from './Description';
+import Tooltip from './Tooltip';
 
 function Checkbox(props) {
   const {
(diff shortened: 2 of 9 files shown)


barmac:
I will look into this on Monday.

assets/properties-panel.css
```diff
@@ -1112,4 +1121,57 @@ textarea.bio-properties-panel-input {
...
   margin-left: auto;
+}
+
+.bio-properties-panel-entry:has(.bio-properties-panel-tooltip-wrapper) .bio-properties-panel-label {
+  text-decoration: underline;
+  text-decoration-style: dotted;
+  text-underline-offset: 2px;
+}
+
+.bio-properties-panel-entry:has(.bio-properties-panel-tooltip) {
```
barmac, line 1132: I love `:has`, but it's not supported in Firefox: https://caniuse.com/?search=%3Ahas

barmac (changes requested):
On Firefox, the styles are not quite there yet:

image

Also, some interactions feel glitchy to me:

https://github.com/bpmn-io/properties-panel/assets/28307541/0ddd4788-dd94-4b11-82d1-871b33e47aec

* When tooltip label is focussed, I'd expect the tooltip to remain open no matter where the cursor is. Then, it should close on blur.
* The click interaction seems to be not quite there. I'd expect to either keep the tooltip open or close it, but never flash-close to open again.

src/components/entries/Checkbox.js
```diff
@@ -71,6 +77,7 @@ function Checkbox(props) {
  * @param {Function} props.setValue
  * @param {Function} props.onFocus
  * @param {Function} props.onBlur
+ * @param {String|import('preact').Component} props.tooltip
```
barmac, line 80: Usually primitive types are lowercase, cf. https://jsdoc.app/tags-type.html
```suggestion
 * @param {string|import('preact').Component} props.tooltip
```

src/components/entries/Tooltip.js
```diff
@@ -0,0 +1,123 @@
...
+              {value}
+            </div>
+            <div class="bio-properties-panel-tooltip-arrow" />
+          </div>
+          : null
+        }
+      </div>
+    );
+  } else {
+    return props.children;
```
barmac, line 105: I don't know if this affects the performance a lot, but currently we register all the hooks even for non-tooltipped entries. Perhaps a cleaner implementation could use two elements:

```jsx
export function TooltipWrapper(props) {
  if (!props.tooltip) {
    return props.children;
  }

  return <Tooltip { ...props } />
}

function Tooltip(props) {
  /* there goes implementation */
}
```

src/components/entries/Tooltip.js
```diff
@@ -0,0 +1,123 @@
...
+
+    const hideTooltipViaEscape = (e) => {
+      e.code === 'Escape' && hideTooltip();
+    };
+
+    current.addEventListener('mouseover', showTooltip);
+    current.addEventListener('focusin', showTooltip);
+    current.addEventListener('keydown', hideTooltipViaEscape);
+    document.addEventListener('focusout', hideFocusedTooltip);
+    document.addEventListener('mousemove', hideHoveredTooltip);
```
barmac, line 71: What's the reason we use DOM elements and not preact wrappers (onMouseMove etc.)?

barmac: I think I know why we do it for `document` listeners but not for the element.

barmac: Also: `mouseover` vs. `mouseenter`, `focusout` vs. `blur`?

barmac:
I left multiple comments due to high complexity of the implemented feature. Overall, I am happy with the implementation, and I believe we have already accomplished a lot. The comments are more about the last mile and polishing the UX which is already good.
Feel free to reach out to me if anything in the comments is unclear. I am also happy to work together to get this great feature merged :)

barmac:
Last one: Perhaps we could also test nested component scenario, as the example branch linked in the PR contains string tooltips only. Would be cool to see how it behaves with focussable elements, e.g. links.

smbea:
> Last one: Perhaps we could also test nested component scenario, as the example branch linked in the PR contains string tooltips only. Would be cool to see how it behaves with focussable elements, e.g. links.

Regarding this one, you can already test the timer props which have a more complex tooltip. But I can add more for the next round

barmac:
Thanks, I missed that!

assets/properties-panel.css
smbea: Forgot about that 🙈

src/components/entries/Tooltip.js
```diff
@@ -0,0 +1,123 @@
...
+
+  const wrapperRef = useRef(null);
+  const tooltipRef = useRef(null);
+
+  const showTooltip = async () => {
+
+    setTimeout(() => {
+      setShow(true);
+    }, 200);
+  };
```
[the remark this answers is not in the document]
smbea: I opted for clearing the timeout on mouse leave because there I didn't find an easy way to check the mouse position to see if it's within bounds and `element.matches(":hover")` made testing a bit more difficult. The `clearTimeout` solution seems to do the trick

src/components/entries/Tooltip.js
smbea: Good idea

src/components/entries/Tooltip.js
smbea: `focusout` over `blur` is because `focusout` bubbles, while `blur` does not. So we can add it the document and listen to any `focusout` events.

`mouseover` vs. `mouseenter` there wasn't really a reason. I changed it and works just fine

barmac:
I am looking into this right now 👍

barmac:
> When tooltip label is focussed, I'd expect the tooltip to remain open no matter where the cursor is. Then, it should close on blur.

Let's fix this too:


https://github.com/bpmn-io/properties-panel/assets/28307541/d2b95365-0884-4b3d-b252-90818f9fa124

barmac:
One thing I had not noticed before: User should be able to select text in the tooltip. Perhaps we should `preventDefault` the clicks from inside of the tooltip.

https://github.com/bpmn-io/properties-panel/assets/28307541/6178a6f5-340a-4ba7-b362-a04835aeeefb

Feel free to have a separate issue for this if you think it's not critical.

barmac:
It looks good on Firefox now! 🏅

src/components/entries/Tooltip.js
```diff
@@ -0,0 +1,138 @@
...
+
+    document.addEventListener('focusout', hideFocusedTooltip);
+    document.addEventListener('mousemove', hideHoveredTooltip);
+
+    return () => {
+
+      document.removeEventListener('mousemove', hideHoveredTooltip);
+      document.removeEventListener('focusout', hideFocusedTooltip);
+    };
+  }, [ wrapperRef, visible ]);
```
barmac, line 90: Minor: Refs are stable by design; we should rather care about `wrapperRef.current` in the dependencies.

barmac:
Feel free to merge this after you look into the remaining comments. I believe it's good enough to be merged without additional round 👏

smbea:
> When tooltip label is focussed, I'd expect the tooltip to remain open no matter where the cursor is. Then, it should close on blur.

Oops I had missed this one. Actually I discussed this with Niklas yesterday about this and we discussed that while generally focus has priority, it may be confusing in this case but neither of us had a very strong opinion. Since you also mentioned this, I will keep it shown while in focus and we can deal with it if there are any "complaints".

> One thing I had not noticed before: User should be able to select text in the tooltip. Perhaps we should preventDefault the clicks from inside of the tooltip.

Fixed

assets/properties-panel.css
```diff
@@ -406,6 +414,7 @@
   display: block;
   font-size: var(--text-size-small);
   margin: 2px 0 1px;
+  width: fit-content;
```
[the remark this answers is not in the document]
smbea: I was trying out the position if we had to open to the right, for example. The width of the label needed to be "true" in that case. But it's not relevant right now. I removed it

PR #265, opened 2023-07-27
Draggable popup component + FEEL popup editor

pinussilvestrus:
Related to https://github.com/camunda/team-hto/issues/262

----

This implements
* a draggable popup component
* a popup editor component for FEEL and templating

Notes for the reviewer:
* I decided to build the FEEL popup editor as a singleton to ensure we only have one popup opened at a time. This is exclusive to FEEL and not something related to popups in general.
* I tried out [`focus-trap`](https://github.com/focus-trap/focus-trap) for the popup component and really liked it for that purpose. The bundle size is okay from my point of view, but I'm open to any feedback here!
* I re-used the [dragger utility](https://github.com/camunda/camunda-modeler/blob/develop/client/src/util/dom/dragger.js) from the camunda-modeler as it does what it should
* There is a demo available for forms via: https://demo-popup-editor-demo--camunda-form-playground.netlify.app/
* In case you want to try it out for BPMN, link this branch to `bpmn-js-properties-panel` locally and execute `npm run start`

Please let me know when you need anything else or whether we should go through the changes together 👍 

image


christian-konrad:
@pinussilvestrus I think we wanted to increase the min-width; it still does not cover our example text here and requires scrolling:

Bildschirmfoto 2023-07-27 um 14 04 25

Bildschirmfoto 2023-07-27 um 14 04 44

It has just 11 chars more width, but I believe it should have at least approx 50 chars more space (700px instead of 400)

Bildschirmfoto 2023-07-27 um 14 06 18

Additional small thing (nothing to care about now but nice-to-have follow-up for spring cleaning): The editor can disappear on viewport resizing, but it's easy to recover by just reopening it:

https://github.com/bpmn-io/properties-panel/assets/56470362/f48fd225-bb97-4beb-847c-06824870c9a0

pinussilvestrus:
> It has just 11 chars more width, but I believe it should have at least approx 50 chars more space (700px instead of 400)

for some reason, I missed this one 😅

christian-konrad:
The button has system font instead of IBM Plex Sans due to CSS execution order:

image

image

RomanKostka (changes requested):
Nice!

i would propose to have a background for the icon so that it stands out clearly.
Currently it overlays with the type is not discoverable:
image
image


We could mititgate this by giving it a white background-color:
image

smbea:
@pinussilvestrus What does `focus-trap` help us accomplish here?

pinussilvestrus:
We use focus traps to keep the focus inside the popup, so when pressing `TAB`, the focus won't ever go outside of it, which makes it saver to use via keyboard. We do the same for example in the Camunda Modeler [with modals and overlays](https://github.com/camunda/camunda-modeler/blob/develop/client/src/shared/ui/trap/FocusTrap.js).

The focus-trap library is pretty configurable, extensible and comes with restoring focus after close, something [we'd have to implement on our own otherwise ](https://github.com/camunda/camunda-modeler/blob/develop/client/src/shared/ui/trap/CloseTrap.js).

smbea:
@pinussilvestrus Makes sense, thanks for explaining.

After the popup is open and you select another form element, the popup remains open and uses the existing input to edit the corresponding property in another form element. I find this a little confusing. Was this behaviour discussed?

https://github.com/bpmn-io/properties-panel/assets/25825387/8f8e7630-c414-4821-a821-39f710e99b6a

In a related note (and maybe this is an edge case we don't care too much about): after changing the focus with mouse (eg selecting another form element), then the focus isn't trapped in the popup anymore

https://github.com/bpmn-io/properties-panel/assets/25825387/937c97de-16df-450c-ac9c-5fc704fb4d0b

christian-konrad:
@smbea @pinussilvestrus 

> After the popup is open and you select another form element, the popup remains open and uses the existing input to edit the corresponding property in another form element. I find this a little confusing. Was this behaviour discussed?


This is not intended, and can lead to strange problems. We should fix it

christian-konrad (changes requested):
We must fix the problem that @smbea reported (clicking another component while the editor is open causes changes to the value of this component's propertsy)

christian-konrad:
Found one issue (if too complex to handle quickly, let's move it to a separate issue in the backlog):

You can still copy-paste FEEL expressions to the source field. It will not paste the expression, but toggles the FEEL editor in the source field

Not a big issue, easy to recover

https://github.com/bpmn-io/properties-panel/assets/56470362/065d6381-549f-4e1e-ad41-c8da12c37d5b

pinussilvestrus:
> After the popup is open and you select another form element, the popup remains open and uses the existing input to edit the corresponding property in another form element. I find this a little confusing. Was this behaviour discussed?

Nice catch! I wonder what the expected behavior is here: do we close the popup or keep the context element in the popup on change?

christian-konrad:
@pinussilvestrus 

> keep the context element in the popup on change?

We should keep it open and editing the previously selected component - in case that's technically feasible.
That allows me to inspect other components and copy variables or code, for example

pinussilvestrus:
> @pinussilvestrus
> 
> > keep the context element in the popup on change?
> 
> We should keep it open and editing the previously selected component - in case that's technically feasible. That allows me to inspect other components and copy variables or code, for example

@christian-konrad I had a look at this during my internet absence (🥲) and I think this is not so easily possible. The properties panel follows the pattern of _always editing the currently selected element_. I'd assume that breaking this up for the popup editor would mean some refactorings. 

I believe closing the popup on selecting another element mitigates at least the broken behavior; I opened https://github.com/bpmn-io/properties-panel/issues/270 to follow up with a more sophisticated approach.

When you think we should give this a higher priority, let's investigate further 👍

pinussilvestrus:
I fixed the following reported issues ✅ 

> i would propose to have a background for the icon so that it stands out clearly.

> You can still copy-paste FEEL expressions to the source field. It will not paste the expression, but toggles the FEEL editor in the source field

> after changing the focus with mouse (eg selecting another form element), then the focus isn't trapped in the popup anymore

smbea:
I like how it's looking now. No concerns implementation wise. Good to go on my end!

RomanKostka (approved):
nice!
all good from my point of view

christian-konrad:
@pinussilvestrus thanks for opening the follow-up issue! I will create a dedicated epic for it, since it covers it's own user problem: editing one property while inspecting others at the same time. And put it on the backburner for now.

Glad to see this merged!

PR #273, opened 2023-08-02
fix(tooltip): content selectable

### LINKED ISSUE #272: Tooltip content isn't selectable
### Describe the Bug

Tooltip content isn't selectable

### Steps to Reproduce

1. open tooltip with text/code
2. try to select text

### Expected Behavior

Content should be selectable

### Environment

- Host (Browser/Node version), if applicable: [e.g. MS Edge 18, Chrome 69, Node 10 LTS]
- OS: [e.g. Windows 7]
- Library version: [e.g. 2.0.0]

smbea:
Closes #272

We added a `preventDefault` to fix a group headers issue, and that  resulted in blocking the text selection. I believe this fixes it


pinussilvestrus:
Do you plan to have this fix in the current release already? I'm asking because I already released the popup editor via a minor version, so bumping the properties panel would mean we introduce this everywhere.

Not a problem, just wanted to make sure we are aware of this 👍

barmac:
Thanks for a quick fix. How would I test this? Can you share what the root cause of the problem is?

smbea:
> Thanks for a quick fix. How would I test this?

You can test it by liking this branch in `bpmn-js-properties-panel`. I recommend trying it out with the timer props because the tooltip has more content.

>  Could you share what the root cause of the problem is?

Like I mentioned above, [we added a `preventDefault` on `mouseDown` to the container div  ](https://github.com/bpmn-io/properties-panel/commit/f0368bb10352a02d540956447eea8915c3a8d25c#diff-8d768f7617a57281431fa0aa72faff214be921ea8824c97257eb0d699b5c80daR111)  to ensure group headers tooltips wouldn't stay always visible after clicking on them to open. This made it so that when trying to select the content, the default selection behaviour didn't happen. So I stopped propagation of the events within the tooltip content div. Happy to explain further via call if my explanation was confusing

barmac:
So if I get it correctly, we want to keep the interaction within tooltip. Is there any need for prevent default in that case?

barmac:
Also, it would be cool to have automatic tests for both `renderPortal` and direct render case. Thus, we can easily verify the correct behavior, and more important catch it when it breaks again.

smbea:
> So if I get it correctly, we want to keep the interaction within tooltip. Is there any need for prevent default in that case?

The prevent default prevents the trigger element (group header) to become focused via mouse, thus keeping the tooltip open. We only want that to happen via keyboard navigation.

> Also, it would be cool to have automatic tests for both renderPortal and direct render case. Thus, we can easily verify the correct behavior, and more important catch it when it breaks again.

I can add test cases for `createPortal`

smbea:
There is actually a test for rendering via portal

I haven't been able to add a test case that reproduces the content selectable or the focus issue I mentioned was fixed with prevent default.

smbea:
When trying to fix #272 and https://github.com/camunda/camunda-modeler/issues/3756, I ended up breaking one while fixing the other. This is because preventing default actions in some places (like focus on `mouseDown`) also ends up preventing wanted actions (like selecting context).

In order to not end up with super confusing code trying to gracefully handle these events, I decided to clean up how we open/persist tooltips based on focus.

I also added some test cases which should help enforce this, but it's hard to test these things for every case so feel free to  give it a good user test.

marstamm (approved):
I can confirm that this fixes the selection issue in a browser environment. Good job cleaning up the opening handling.

As expected, in the Camunda Modeler, selection is still blocked for the whole modeler container

Recording 2023-08-14 at 10 48 37

PR #278, opened 2023-08-29
chore(build): publish css in `dist/assets`

marstamm:


assets/properties-panel.css
```diff
@@ -1,1435 +1,4 @@
...
-}
-
-.bio-properties-panel-feel-popup .cm-gutter {
-  min-width: 32px;
-}
-
-.bio-properties-panel-feel-popup .cm-gutters .cm-lineNumbers .cm-gutterElement {
-  text-align: center;
-}
+@import '../dist/assets/properties-panel.css'
```
marstamm, line 4: We keep the `assets` file with a link for backwards compatibility

nikku: Do we have to? I guess we could simply drop it, as this was not supposed to be consumed in the past (was consumed via `*-properties-panel` until now).

nikku: :+1:

Issue #277, opened 2023-08-25
Pop container can easily cause the entire page to overflow.Can you add fill to currentcolor for the icon to facilitate color setting


zxuanhong:
### Describe the Bug

1. The bullet frame should not be mounted under the root body.This can cause penetration out of the container, resulting in a page overflow
2. The top left icon color should be modifiable to fit the dark mode

<img width="1687" alt="image" src="https://github.com/bpmn-io/properties-panel/assets/62740686/29491b52-f9a5-4424-b4fa-5415b64b1dc4">

### Environment

- Host (Browser/Node version), if applicable: [e.g. MS Edge 18, Chrome 69, Node 10 LTS] chrome 110
- OS: [e.g. Windows 7] mac
- Library version: [e.g. 2.0.0]  3.3.2

nikku:
CC @pinussilvestrus. Maybe you can have a look?

zxuanhong:
And how do I disable this pop-up

christian-konrad:
@zxuanhong you can easily change colors of the drag icon with css filters

zxuanhong:
@christian-konrad I think it's easier to set up with currentColor

pinussilvestrus:
> The bullet frame should not be mounted under the root body.This can cause penetration out of the container, resulting in a page overflow

With https://github.com/bpmn-io/properties-panel/pull/280 we aim to add an option to configure the parent container of the popup editor.

zxuanhong:
@pinussilvestrus I only used bpmn js, so  need to adjust the bpmn-js-properties-panel component configuration

pinussilvestrus:
@zxuanhong this option will be present via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/970 in bpmn-js-properties-panel as well

zxuanhong:
@pinussilvestrus Thank you very much. I think there should be no problem only to set up, so I will close the current issue

PR #280, opened 2023-09-07
feat(feel-popup): allow defining popup container

pinussilvestrus:
Related to https://github.com/bpmn-io/properties-panel/issues/277

This allows to set the parent container of the FEEL popup via a property. The default keeps `document.body`.

For context: We had problems in the Web Modeler properly cleaning up the popups when navigating outside of the editor (cf. https://camunda.slack.com/archives/C046XS3GGCR/p1693400946971909). For example, in the forms editor, we can set the container to the Form Playground Root.

Cf. to the implementation in form-js as an example of how to use this.
* https://github.com/bpmn-io/form-js/pull/795
* https://github.com/camunda/form-playground/pull/101 ([Demo](https://feel-popup-root--camunda-form-playground.netlify.app/))

image


nikku (approved):
Makes a lot of sense + works as in other components.

PR #285, opened 2023-09-15
fix(FEEL): reported error switching

smbea:
Related to https://github.com/camunda/camunda-modeler/issues/3845

### DIFF
--- a/src/components/entries/FEEL/Feel.js
+++ b/src/components/entries/FEEL/Feel.js
@@ -566,7 +566,7 @@ export default function FeelEntry(props) {
 
   const temporaryError = useError(id);
 
-  const error = localError || temporaryError || validationError;
+  const error = temporaryError || localError || validationError;
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


nikku:
Do you remember why we had `local > global` in the past? And can you make a good call for `global > local`? I'm happy to merge this, we just want to ensure we know why this is a good direction.

smbea:
> Do you remember why we had local > global in the past?

I don't really know. I guess it's so the user can get the most context: a syntax error is more specific than just "not valid"

> And can you make a good call for global > local?

I vouch for this because you can still get the syntax error by hovering the highlighted area.

image

nikku:
Thanks. Makes sense to me, along with prior discussions in https://github.com/camunda/camunda-modeler/issues/3845#issuecomment-1721081797 and following.

If there is a global (in the larger context) error provided, we'll always show that _over_ the local error.

nikku (approved):

nikku:
Thinking about it `global` should rather be called `externally provided` to make that clear. Not required for this fix though.

PR #287, opened 2023-09-22
Feel Popup event support

pinussilvestrus:
Related to https://github.com/bpmn-io/bpmn-js-properties-panel/issues/974

This provides the capability to hook into the feel popup lifecycle as
* `feelPopup.opened`
* `feelPopup.closed`
* `feelPopup._open`
* `feelPopup._close`
* `feelPopup._isOpen`

Furthermore, this adds a feature service that integrators can use to interact with the popup. Cf. to this example [on how to use it](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/975). I didn't add particular test coverage for this module in this project similar to `debounceInput` as I saw we do not have a proper didi-like test setup in the core framework.

```js
const feelPopup = bpmnjs.get('feelPopup');

console.log('is open', feelPopup._isOpen());

feelPopup.close();

feelPopup.open(someElement, { type: 'feel' }, elementToRestoreFocusAfterClose);
```

Happy to receive your feedback ❤️ @bpmn-io/modeling-dev @bpmn-io/hto-dev

### DIFF
--- a/src/components/entries/FEEL/FeelPopup.js
+++ b/src/components/entries/FEEL/FeelPopup.js
@@ -1,4 +1,5 @@
 import {
+  useCallback,
   useEffect,
   useRef,
   useState
@@ -24,6 +25,7 @@ export const FEEL_POPUP_HEIGHT = 250;
 export default function FEELPopupRoot(props) {
   const {
     element,
+    eventBus,
     popupContainer
   } = props;
 
@@ -34,16 +36,26 @@ export default function FEELPopupRoot(props) {
   const [ source, setSource ] = useState(null);
   const [ sourceElement, setSourceElement ] = useState(null);
 
-  const handleOpen = (key, config, _sourceElement) => {
-    setSource(key);
+  const emit = (type, context) => {
+    eventBus.fire('feelPopup.' + type, context);
+  };
+
+  const isOpen = useCallback(() => {
+    return !!open;
+  }, [ open ]);
+
+  const handleOpen = (entryId, config, _sourceElement) => {
+    setSource(entryId);
     setPopupConfig(config);
     setOpen(true);
     setSourceElement(_sourceElement);
+    emit('opened');
   };
 
   const handleClose = () => {
     setOpen(false);
     setSource(null);
+    emit('closed');
   };
 
   const feelPopupContext = {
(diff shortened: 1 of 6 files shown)


src/components/entries/FEEL/FeelPopup.js
```diff
@@ -54,11 +66,48 @@ export default function FEELPopupRoot(props) {
...
+    eventBus.on('feelPopup.close', handleClose);
+    eventBus.on('feelPopup.open', handlePopupOpen);
+    eventBus.on('feelPopup.isOpen', handleIsOpen);
+
+    return () => {
+      if (!eventBus) {
+        return;
+      }
+
+      eventBus.off('feelPopup.close', handleClose);
```
nikku, line 104: I suggest that we separate external events you're supposed to use (`feelPopup.opened`) from internal ones (i.e. `_open`, `_isOpen`, ...).

Other than that this is a great addition. Shows how we can make core react components accessible to the outside, behind a clear, known, contract. :muscle:

pinussilvestrus: Good suggestion, makes a lot of sense! Updated it accordingly 👍

src/components/entries/FEEL/FeelPopup.js
```diff
@@ -34,16 +36,26 @@ export default function FEELPopupRoot(props) {
   const [ source, setSource ] = useState(null);
   const [ sourceElement, setSourceElement ] = useState(null);
 
-  const handleOpen = (key, config, _sourceElement) => {
-    setSource(key);
+  const emit = (type, context) => {
+    eventBus.fire('feelPopup.' + type, context);
```
nikku, line 40: If we want to keep backwards compatibility then we'd need to make this optional.

```suggestion
    eventBus && eventBus.fire('feelPopup.' + type, context);
```

src/components/entries/FEEL/FeelPopup.js
```diff
@@ -24,6 +25,7 @@ export const FEEL_POPUP_HEIGHT = 250;
 export default function FEELPopupRoot(props) {
   const {
     element,
+    eventBus,
```
nikku, line 28:
```suggestion
    eventBus = { fire() {}, on() {} }
```

nikku: Alternative :)

nikku (dismissed):
This looks great. One last question: Do we need to make `eventBus` an optional dependency at the core?

pinussilvestrus:
I liked the null-object pattern usage here, so I used that via https://github.com/bpmn-io/properties-panel/pull/287/commits/0abe5d777de35f04401122bffa76276bb95c717b. 🏆 

I believe that is okay, as `eventBus` is optional in the parent `PropertiesPanel` component as well (at least from the JSDocs 🙂)

nikku (approved):
Great stuff.

pinussilvestrus:
I'd take over releasing these bits and open up https://github.com/bpmn-io/bpmn-js-properties-panel/pull/975 if that's okay to you.

nikku:
That's great, thank you :)

PR #293, opened 2023-09-25
Fix data marker update

smbea:
Related to https://github.com/camunda/camunda-modeler/issues/3815

As per https://github.com/camunda/camunda-modeler/issues/3815#issuecomment-1733038161

### DIFF
--- a/src/components/Group.js
+++ b/src/components/Group.js
@@ -58,6 +58,8 @@ export default function Group(props) {
 
   // set edited state depending on all entries
   useEffect(() => {
+
+    // TODO(@barmac): replace with CSS when `:has()` is supported in all major browsers, or rewrite as in https://github.com/camunda/camunda-modeler/issues/3815#issuecomment-1733038161
     const scheduled = requestAnimationFrame(() => {
       const hasOneEditedEntry = entries.find(entry => {
         const {
--- a/src/components/Group.js
+++ b/src/components/Group.js
@@ -58,25 +58,29 @@ export default function Group(props) {
 
   // set edited state depending on all entries
   useEffect(() => {
-    const hasOneEditedEntry = entries.find(entry => {
-      const {
-        id,
-        isEdited
-      } = entry;
+    const scheduled = requestAnimationFrame(() => {
+      const hasOneEditedEntry = entries.find(entry => {
+        const {
+          id,
+          isEdited
+        } = entry;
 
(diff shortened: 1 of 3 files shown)


test/spec/components/Group.spec.js
```diff
@@ -43,6 +43,11 @@ describe('<Group>', function() {
     container = TestContainer.get(this);
   });
 
+  beforeEach(() => {
+    const requestAnimationFrameSpy = sinon.fake.returns(cb => cb());
+    global.window.requestAnimationFrame = requestAnimationFrameSpy();
+  });
```
barmac, line 49: The spy is not cleaned up properly, we should use sth like:

```javascript
afterEach(() => global.window.requestAnimationFrame.restore());
```

src/components/Group.js
```diff
@@ -58,25 +58,29 @@ export default function Group(props) {
 
   // set edited state depending on all entries
   useEffect(() => {
-    const hasOneEditedEntry = entries.find(entry => {
-      const {
-        id,
-        isEdited
-      } = entry;
+    const scheduled = requestAnimationFrame(() => {
```
barmac, line 61:
```suggestion
    // TODO(@barmac): replace with CSS when `:has()` is supported in all major browsers, or rewrite as in https://github.com/camunda/camunda-modeler/issues/3815#issuecomment-1733038161
    const scheduled = requestAnimationFrame(() => {
```

test/spec/components/Group.spec.js
```diff
@@ -43,6 +43,11 @@ describe('<Group>', function() {
...
+
```
barmac, line 50:
```suggestion

  afterEach(() => global.window.requestAnimationFrame.restore());
```

barmac:
I applied my suggestions from code review; after the CI finishes let's merge this!

barmac (approved):

test/spec/components/Group.spec.js
smbea: Seems like this doesn't work: https://github.com/bpmn-io/properties-panel/actions/runs/6300214448/job/17102537862?pr=293#step:6:243

barmac: Indeed, it looks like fakes don't have `restore` even though it's in the types file.

https://sinonjs.org/releases/latest/fakes/

barmac:
Ok now the tests with `restore` work fine locally, and should also pass on CI.

Issue #290, opened 2023-09-25
For the FEEL popup editor, provide more detailed focused state


marstamm:
### Why should we do it?

In the Desktop Modeler, we have global keyboard shortcuts that trigger independent of the handling in the DOM tree. For a good editing experience, these have to be disabled when focusing on a text editor.

### What should we do?

Provide `focusIn` and `focusOut` events.

alternative: have the focusIn and focusOut events added from the integrating editor, [like we do for properties panel focus events](https://github.com/camunda/camunda-modeler/blob/29f0379b770dea28dfe5dba3166322d133656a8e/client/src/app/tabs/bpmn/modeler/features/properties-panel-keyboard-bindings/PropertiesPanelKeyboardBindings.js). For this, it would be helpful if the `.opened` event contains the DOM element of the popup so additional event listeners can be attached.

related to https://github.com/camunda/camunda-modeler/issues/3877#issuecomment-1733199580

nikku:
I'm in favor of doing this, cf. https://github.com/bpmn-io/properties-panel/pull/294#issuecomment-1733778560.

marstamm:
Closed with #294

PR #294, opened 2023-09-25
feat(FEEL): adjust lifecycle events to match conventions

marstamm:
Adds new lifecycle events:

  - `feelPopup.open` - fired before the popup is mounted
  - `feelPopup.opened` - fired after the popup is mounted. Event context contains the DOM node of the popup
  - `feelPopup.close` - fired before the popup is unmounted. Event context contains the DOM node of the popup.
  - `feelPopup.closed` - fired after the popup is unmounted
 
---

Context: 
This might be a bit over-engineered. We basically need 2 events to allow us to implement `focusin` and `focusout` as a separate component as described in #290. These are:
 - `opened`, fired after mounting the DOM element
 - `close`, fired before unmounting

I also added the corresponding `open` (before mount) and `closed` (after unmount) events for completeness, even if we do not use them. I'm not sure on this one, as it adds a bit more complexity than we don't need.
The `opened` and `close` events also provide the DOM node of the popup which allows us to add additional event handlers outside of the properties panel code. I tested locally that this allows us to properly handle focus state in the desktop modeler.

Previously, `opened` was fired before the Component was mounted and `closed` before the element was unmounted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,7 @@ All notable changes to [`@bpmn-io/properties-panel`](https://github.com/bpmn-io/
 
 ___Note:__ Yet to be released changes appear here._
 
-* `FEAT`: improve FEEL popup lifecycle events:
-  - `feelPopup.open` - fired before the popup is mounted
-  - `feelPopup.opened` - fired before the popup is unmounted. Event context contains the DOM node of the popup
-  - `feelPopup.close`- fired before the popup is unmounted. Event context contains the DOM node of the popup.
-  - `feelPopup.closed`- fired after the popup is unmounted
-
+* `FEAT`: improve FEEL popup lifecycle events ([#294](https://github.com/bpmn-io/properties-panel/pull/294))
 
 ## 3.7.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to [`@bpmn-io/properties-panel`](https://github.com/bpmn-io/
 
 ___Note:__ Yet to be released changes appear here._
 
+* `FEAT`: improve FEEL popup lifecycle events:
+  - `feelPopup.open` - fired before the popup is mounted
+  - `feelPopup.opened` - fired before the popup is unmounted. Event context contains the DOM node of the popup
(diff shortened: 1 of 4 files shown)


nikku:
Maybe silly question: If what we need is `focusin` and `focusout`, why don't we implement these events?

marstamm:
For the properties-panel, we also handle focusin/out in the [modeler itself](https://github.com/camunda/camunda-modeler/blob/29f0379b770dea28dfe5dba3166322d133656a8e/client/src/app/tabs/bpmn/modeler/features/properties-panel-keyboard-bindings/PropertiesPanelKeyboardBindings.js), so I decided to go in a similar direction. I also think it looks a bit cleaner when having this functionality optionally as a bpmn-js extension. I guess what it comes down to is if we want to have this functionality for everyone, even when we probably only need it in the modeler.

This PR simply ensures that the events we fire are meaningful enough to allow integrating libraries to extend the functionality by themselves :)

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ All notable changes to [`@bpmn-io/properties-panel`](https://github.com/bpmn-io/
 
 ___Note:__ Yet to be released changes appear here._
 
+* `FEAT`: improve FEEL popup lifecycle events:
```
nikku, line 9: We could consider to make this less verbose, and simply link to the (comprehensive) PR description. We want to link to a PR anyway.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ All notable changes to [`@bpmn-io/properties-panel`](https://github.com/bpmn-io/
...
+  - `feelPopup.open` - fired before the popup is mounted
+  - `feelPopup.opened` - fired before the popup is unmounted. Event context contains the DOM node of the popup
```
barmac, line 11:
```suggestion
  - `feelPopup.opened` - fired after the popup is mounted. Event context contains the DOM node of the popup
```

barmac: Or remove it from the changelog as @nikku suggested.

marstamm: Good catch. I'll remove it from the changelog and paste it into the description of this PR

nikku (approved):

PR #296, opened 2023-09-29
feat(Popup): allow `PopupContainer` to be a CSS selector

### LINKED ISSUE #291: For the FEEL popup editor, allow `feelPopupContainer` to be a CSS selector
### What should we do?

Allow the config option `config.propertiesPanel.feelPopupContainer` to optionally be a CSS selector

### Why should we do it?

This is in line with other places where we require a container, like the bpmn-js container or the `feelTooltipContainer`.
This is required because we might [bootstrap the editor cached without the container present](https://github.com/camunda/camunda-modeler/blob/7297f1607d2d1ed6fad4aa24c38fb56b17f31c96/client/src/app/tabs/cloud-bpmn/BpmnEditor.js#L840). In these cases, the container should be determined at render time (cf. https://github.com/camunda/camunda-modeler/issues/3877#issuecomment-1733198745)

marstamm:
closes #291

### DIFF
--- a/src/components/Popup.js
+++ b/src/components/Popup.js
@@ -1,16 +1,16 @@
 import { createPortal, forwardRef } from 'preact/compat';
 
-import { useEffect, useRef } from 'preact/hooks';
+import { useEffect, useMemo, useRef } from 'preact/hooks';
 
 import classNames from 'classnames';
+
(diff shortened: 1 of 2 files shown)


barmac (approved):
LGTM

PR #300, opened 2023-10-11
fix(tooltips): remove pre code double borders

Skaiir:
It's pretty standard for code blocks to use both `<pre>` and `<code>` html tags together. I have removed the double border which arises when using this setup, while keeping the border when they are used individually. 

grafik


barmac (approved):
LGTM

PR #301, opened 2023-10-12
Fix: Buttons without types

### LINKED ISSUE #299: Consider adding type="button" to <button> elements
### Is your feature request related to a problem? Please describe.

I had a long fight with a properties panel where clicking e.g. the buttons to expand a group (in a header) would cause the browser to navigate. I finally tracked it down to a `<form>` ancestor getting submitted whenever by the `<button>` clicks.

### Describe the solution you'd like

The default action for `<button>`s without a `type` attribute is to submit the closest ancestor form. I don't think that was the intention with any of your buttons. If you add a `type="button"` attribute to your buttons you won't get this behaviour. See e.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button.

### Describe alternatives you've considered

I added a `onsubmit="return false"` attribute to my form which solved the issue for me (since it is never supposed to be submitted anyway).

jonathanbp:
This PR adds `type="button"` to all `<button>` elements. Could maybe be done nicer with a custom `<Button>` component at some point.

Closes #299.


nikku:
Thanks for your contribution!

PR #304, opened 2023-10-13
chore: improve build scripts

nikku:
* introduce `clean` script
* re-use `build` during `build:watch`

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,9 +10,10 @@
     "preact"
   ],
   "scripts": {
-    "all": "run-s lint build test",
-    "build": "del-cli preact dist && rollup -c",
-    "build:watch": "rollup -c --watch",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -10,9 +10,10 @@
     "preact"
   ],
   "scripts": {
-    "all": "run-s lint build test",
-    "build": "del-cli preact dist && rollup -c",
```
barmac, line 14: So we don't clean up preact anymore?

nikku: My bad, fixed via https://github.com/bpmn-io/properties-panel/pull/304/commits/961825cf25d7b7e10f6e40d5e3c9e1b78d3a07f2.

barmac (approved):

PR #305, opened 2023-10-13
Revert `mjs` rename

nikku:
This causes [unintended side-effects](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/985#issuecomment-1761162479) in module bundling.


barmac (approved):

nikku:
Thanks for the swift review!

PR #306, opened 2023-10-13
Unbundle `preact` dependency

nikku:
This ensures `preact` is not inlined anymore.

---

Child of https://github.com/bpmn-io/internal-docs/issues/819


Skaiir:
@nikku Resurrecting this a little bit to make a point. Properties panel is a component library. And the way component libraries should expose preact or react is well agreed to be as a peer dependency, not a direct dependency.

This is something that is quite well agreed in the common consciousness ([stack overflow](https://stackoverflow.com/questions/30451556/what-is-the-correct-way-of-adding-react-as-a-dependency-in-the-package-json), AIs, blogs, other component libraries like carbon).

Why we can do it: properties panel exports JSX directly, it doesn't have the responsibility of rendering. Since it's designed to be rendered as part of an upstream library, there will never be a scenario where a consumer doesn't itself have to provide preact.

Why we should do it: peer dependency mismatches will error at package install time, regular dependency mismatches will cause package duplication ->  runtime errors.

I know de-duplication works well but this is about the scenarios where it doesn't.

The way I see it, we should have:

`preact` as peerDependency: `^10.0.0`, provided we verify (through CI) that it is compatible with older preact 10 versions, if it's also compatible with preact 11 we can do  `"^10.0.0 || ^11.0.0"`. We don't need to be opinionated here, widest possible version range is the best, just a tradeoff with how much CI we're willing to throw at it.

`preact` remains a devDependency, and doesn't change, `^10.19.3` is totally fine, whatever version we like to work with.

I should note that as of now, form-js and properties-panel are **not compatible**. We are doing some heavy trickery to bundle the properties panel inside of form-js and aliasing all of the preact deps inside of properties panel (which point to the bundled copy) back to form-js top level. Effectively overriding the copy entirely. It works for now but the moment the properties panel depends on a 1.16+ feature of preact we won't be able to upgrade anymore. (This is what initiated me to resurrect these talks).

(PS the fact that form-js doesn't support preact 10.16.0 and above is also a real issue that needs tackling, but serves as an example of how having a component library opinionated about preact versions can cause upstream issues)

PR #308, opened 2023-10-18
feat(FeelPopup): link to learning resources

marstamm:
related to https://github.com/camunda/camunda-modeler/issues/3929

The "What is FEEL" page will be overhauled as another part of https://github.com/camunda/camunda-modeler/issues/3929

image


src/components/entries/FEEL/FeelPopup.js
```diff
@@ -179,6 +180,14 @@ function FeelPopupComponent(props) {
     return () => emit('close', { domNode: popupRef.current });
   }, []);
 
+  useEffect(() => {
+
+    // Set focus on editor when popup is opened
+    if (editorRef.current) {
+      editorRef.current.focus();
+    }
+  }, [ editorRef ]);
```
marstamm, line 189: Specific focus is required now that the editor is not the first focusable element in the popup.

src/assets/properties-panel.css
```diff
@@ -1271,18 +1271,21 @@ textarea.bio-properties-panel-input {
...
 }
 
 .bio-properties-panel-popup .bio-properties-panel-popup__header {
+  --popup-header-line-hight: 16px;
   background-color: var(--popup-header-background-color);
   margin: 0;
   font-size: 12px;
   font-weight: 400;
-  line-height: 16px;
+  line-height: var(--popup-header-line-hight);
```
nikku, line 1288:
```suggestion
  line-height: var(--popup-header-line-height);
```

src/components/entries/FEEL/FeelPopup.js
nikku: :+1:

src/components/entries/FEEL/FeelPopup.js
```diff
@@ -200,7 +209,22 @@ function FeelPopupComponent(props) {
       <Popup.Title
         title={ title }
         emit={ emit }
-        draggable />
+        draggable>
+        {type === 'feel' && (
+          <a href="https://docs.camunda.io/docs/components/modeler/feel/what-is-feel/" target="_blank" class="bio-properties-panel-feel-popup__title-link">
```
nikku, line 214: I wonder if we want to make this pluggable, or make the popup editor a Camunda-only component (cf. https://github.com/bpmn-io/internal-docs/issues/825). 

To me it does not really make sense to embed such a link inside an agnostic library. Happy to compromise temporarily, of course.

marstamm: I agree, this should not be in an agnostic library. We talked about moving the popout editor into its own component anyway, likely when we want to add more FEEL guidance/data awareness to it. 
So I would not spend the time right now to make the header pluggable

nikku: Then let's keep it as is (we compromise) for the moment.

PR #312, opened 2023-11-30
feat: use popup icon for FEEL editor

### LINKED ISSUE #310: Change the FEEL popup editor icon
### What should we do?

We should change the FEEL popup editor icon to the 'popup' icon.
Popup icon:
image
image

**Additional information:**
* Carbon reference: [icons](https://carbondesignsystem.com/guidelines/icons/library/) ( search for popup )
* [FigmaFile](https://www.figma.com/file/CiTSN1vNETjIAy6W9jUIFp/Form-js-Specifications?type=design&node-id=1623%3A209&mode=design&t=4oYKIfCvFXZ7b7KI-1)

### Why should we do it?

The current icon ( 'launch icon' ) is being used to indicate an external Link.
Launch icon:
image

Therefore we should change the FEEL popup editor icon to the 'popup icon' which makes it clear what to expect.
Popup icon:
image

nikku:
capture ULW8Jk_optimized

Closes #310


smbea:
I just noticed that for bpmn, this icon appears with a white background. This happens now too but I was wondering if we want to keep it. 
Screenshot 2023-11-30 at 09 13 26

This is how it looks for form-js, for example:
Screenshot 2023-11-30 at 09 16 53

RomanKostka:
> I just noticed that for bpmn, this icon appears with a white background. This happens now too but I was wondering if we want to keep it.

Good catch!
@nikku can we display the icon without any background color?

nikku:
Removed white background via https://github.com/bpmn-io/properties-panel/pull/312/commits/00ba974781c1736b29f0bc8e7d61a2019b7ba881.

Looking at the existing code it was explicitly designed that way in the past. I agree that no background is the better option.

PR #318, opened 2023-12-14
pass event.target instead of event to make it also work in the shadow dom 

### LINKED ISSUE #313: TextField throws error when used inside web component (and shadow dom)
### Describe the Bug

TextField (and TextArea) throws error when input changes and the properties panel resides inside a web component.

The scope changes from shadow dom (inside handleInput method instance) to dom (inside handleInputCallback debounce method instance).

Same goes for TextArea.

Checkbox, Select and FEEL editor work as expected.

### Steps to Reproduce

Create web component with code inside:

```javascript
import BaseModeler from 'camunda-bpmn-js/lib/base/Modeler';
import ModelerStyle from "camunda-bpmn-js/dist/assets/base-modeler.css";

const modeler = new BaseModeler({
      container: this.renderRoot.querySelector('#canvas') as HTMLElement,
      propertiesPanel: {
        parent: this.renderRoot.querySelector('#properties') as HTMLElement  
      }
});

await modeler.createDiagram();
```

And then change the name of the process.

### Expected Behavior

[Exception: TypeError: Cannot read properties of undefined (reading 'length')](https://github.com/bpmn-io/properties-panel/blob/8529f45dcf60c8e6e1265d84b3f6aff9703e1fdc/src/components/entries/TextField.js#L38)

Target here is the web component in the dom instead of the input in the shadow dom.

[Inside this method e has the right context and e.target equals the input instance](https://github.com/bpmn-io/properties-panel/blob/8529f45dcf60c8e6e1265d84b3f6aff9703e1fdc/src/components/entries/TextField.js#L42)

### Environment

- Host (Browser/Node version), if applicable: Brave 1.60.125  and node v18.17.1
- macOS Sonoma 14.1.1
- Library version: 3.13.0

arjangeertsema:
Closes #313 

Passed event.target instead of event as parameter to the debounce method.

To fix issue https://github.com/bpmn-io/properties-panel/issues/313

### DIFF
--- a/src/components/entries/NumberField.js
+++ b/src/components/entries/NumberField.js
@@ -35,21 +35,16 @@ export function NumberField(props) {
   const [ localValue, setLocalValue ] = useState(value);
 
   const handleInputCallback = useMemo(() => {
-    return debounce(event => {
+    return debounce((target) => {
 
-      const {
-        validity,
-        value
-      } = event.target;
-
-      if (validity.valid) {
-        onInput(value ? parseFloat(value) : undefined);
+      if (target.validity.valid) {
+        onInput(target.value ? parseFloat(target.value) : undefined);
       }
     });
   }, [ onInput, debounce ]);
 
   const handleInput = e => {
-    handleInputCallback(e);
+    handleInputCallback(e.target);
     setLocalValue(e.target.value);
   };
 
--- a/src/components/entries/Simple.js
+++ b/src/components/entries/Simple.js
@@ -32,11 +32,11 @@ export default function Simple(props) {
   const [ localValue, setLocalValue ] = useState(value);
 
   const handleInputCallback = useMemo(() => {
-    return debounce(({ target }) => setValue(target.value.length ? target.value : undefined));
+    return debounce((target) => setValue(target.value.length ? target.value : undefined));
   }, [ setValue, debounce ]);
 
   const handleInput = e => {
-    handleInputCallback(e);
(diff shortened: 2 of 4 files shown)


nikku:
Thanks for your contribution @arjangeertsema. Looks good.

nikku:
What we _could do_ is to add a dedicated test to verify this works in shadow DOM. This way we can assert it does not break (undeteced) again.

I believe such test would be better put, i.e. into bpmn-js-properties-panel.

arjangeertsema:
@nikku please let me know what the next steps are and what you expect from me. 

Please inform me also about the lead time for these kind of small bug fixes? When can I expect that this bug fix is added to a new release, and when will that release be incorporated in the modeller?

nikku:
> Please inform me also about the lead time for these kind of small bug fixes? When can I expect that this bug fix is added to a new release, and when will that release be incorporated in the modeller?

@arjangeertsema I suggested one improvement on your PR:

> What we could do is to add a dedicated test to verify this works in shadow DOM. This way we can assert it does not break (undeteced) again.

We'll merge the PR without that improvement, but you'll risk getting shadow DOM support broken (again) in the future without an integration test.

nikku (approved):

nikku:
Merged via https://github.com/bpmn-io/properties-panel/commit/0171b6a42e535e6e0028c24fcde492ff4e4757e6.

PR #319, opened 2023-12-15
fix: Enable scrolling on FEEL popup editor

### LINKED ISSUE #316: Scrollbars of the popup are hidden
### Describe the Bug

Long content in the editor can't be properly worked on, as scrollbars are hidden.

There are 2 CSS rules, one in the editor, one in the properties panel, that actively hide the scrollbars.

Bildschirmfoto 2023-12-14 um 15 04 00

Bildschirmfoto 2023-12-14 um 15 04 19

However, if we enable vertical scrollbars, there's an annoying right padding we should get rid of:

Bildschirmfoto 2023-12-14 um 15 10 19

### Steps to Reproduce

1. Open popup editor
2. Add a lot of content
3. Scrolling is hard, no visible scrollbars

### Expected Behavior

The editor shows scrollbars when needed.

vsgoulart:
The scrollbar in the FEEL popup editor was hidden when the text was too long.

https://github.com/bpmn-io/properties-panel/assets/1022916/357374e1-0009-4e01-ab55-7fd2572cc72a

Closes #316

### DIFF
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -1361,6 +1361,10 @@ textarea.bio-properties-panel-input {
   min-width: 100%;
 }
 
+.bio-properties-panel-feel-popup .bio-properties-panel-feel-editor-container .cm-scroller {
+  overflow: auto !important;
+}
+
 .bio-properties-panel-feel-popup .bio-properties-panel-feelers-editor-container {
   width: 100%;
   display: flex;
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -1277,6 +1277,8 @@ textarea.bio-properties-panel-input {
 
 .bio-properties-panel-popup .bio-properties-panel-popup__body:not(:first-child) {
   padding-top: 0;
+  padding-left: 0;
+  padding-right: 0;
 }
 
 .bio-properties-panel-popup .bio-properties-panel-popup__header {
@@ -1382,7 +1384,6 @@ textarea.bio-properties-panel-input {
   width: 100%;
   resize: none;
   padding: 0;
-  margin-left: -12px;
   overflow: hidden;
   overflow-y: auto
 }
--- a/src/components/entries/FEEL/FeelEditor.js
+++ b/src/components/entries/FEEL/FeelEditor.js
@@ -4,7 +4,7 @@ import { forwardRef } from 'preact/compat';
 
 import FeelEditor from '@bpmn-io/feel-editor';
 
-import { lineNumbers } from '@codemirror/view';
+import { EditorView, lineNumbers } from '@codemirror/view';
 
 import { useStaticCallback } from '../../../hooks';
 
@@ -98,7 +98,8 @@ const CodeEditor = forwardRef((props, ref) => {
       value: localValue,
       variables: variables,
       extensions: [
-        ...enableGutters ? [ lineNumbers() ] : []
+        ...enableGutters ? [ lineNumbers() ] : [],
+        EditorView.lineWrapping
       ]
     });
 


src/assets/properties-panel.css
```diff
@@ -1097,10 +1097,6 @@ textarea.bio-properties-panel-input {
   padding: 2px 6px;
 }
 
-.bio-properties-panel-feel-editor-container .cm-scroller {
-  overflow: hidden !important;
```
vsgoulart, line 1101: @marstamm I drilled as far as I could in the history and from what I've seen you added these styles. I couldn't notice anything this was fixing, but maybe I missed something and you remember why this was necessary

nikku: Happy to merge if it also works in feel editors embedded inside the properties panel. There we want the editor to always have full height.

christian-konrad: @nikku is this style affecting the native property panel fields? Then we should probably find a dedicated fix in the popup editor

nikku: > @nikku is this style affecting the native property panel fields? 

I don't know, but I gave this pointer to @vsgoulart to be able to assess the end-to-end impact of removing a prior fix (again).

Skaiir:
### Analysis of the cm-scroller stuff, it's kind of a fascinating evolution that started multiple years ago.

- first version of the feel component shipped with `.cm-scroller: overflow-x: none`, presumably to hide the thick scrollbar that would occur when working with single line expressions
- when feel textareas joined the party (which are now used everywhere), this was extended to `overflow` in both directions, presumably for the same reason (as resizing was introduced)

Example of what it looks like on windows with the scrollbars on:

grafik

- eventually this was scoped to `.bio-properties-panel-feel-editor-container .cm-scroller` to stop the properties panel from interfering with our other codemirror instances
- niklas re-used the container classes to bring over as much of the styling of the properties panel proper into the popup, however this means there's also the disabling scroller stuff in there. 

### Takeaways

So, two ways we can fix this:
- re-enable scrollbars only in the popup, this would mean we'd use `bio-properties-panel-popup` to scope another css rule
- re-enable scrollbars in both, merge this fix

My opinion:
- scrollbars **should** be visible in the properties panel (non-popup), but we should use a subtle custom implementation of them (something like `https://malte-wessel.com/react-custom-scrollbars/`).
- for now, let's just re-enable the scrolling within the popups only
- lets also raise some follow-ups as we see are appropriate

Other observations:
- the positioning of the feel popup editor on open is quite janky, especially when the text content is large
- the popup button prevents resizing of the inner text
- the resizing option might not be serving a purpose anyways, since we autosize
- feel textareas should be height limited, otherwise they become unmanageable
- feel textareas and regular textareas behave differently in general and should be made more similar

Skaiir:
I've sent through my own suggested fix.

philippfromme:
Regarding your other observations, did you check whether there are existing issues for them?

Skaiir:
@philippfromme They don't, at least not directly though some issues are already raised that are similar. For those I just wanted to see what people's thought were but I plan to raise follow-ups

nikku:
Let's clearly separate styles for popup editor and properties panel (cf. https://github.com/bpmn-io/internal-docs/issues/825).

Let's then ship an isolated fix for the popup editor to make scrolling possible there.

The properties panel is meant to be resized to "fit contents".

nikku:
What we could consider is to add [automatic line breaks](https://stackoverflow.com/questions/19756934/codemirror-auto-line-break) (always fit content). This is exactly how text area elements work, too.

marstamm:
> What we could consider is to add [automatic line breaks](https://stackoverflow.com/questions/19756934/codemirror-auto-line-break) (always fit content). This is exactly how text area elements work, too.

I think this makes sense - we evolved from single-line to multi-line editor and did not change the line-break behavior along with it. As we always have a "Text Area style" editor, we should not use the line-break behavior of normal input fields

marstamm:
@Skaiir when testing on properties panel, the scroll-bars are interfering with the Close button. This looks broken and not intended:

image

Are there upstream changes that I am missing which change these styles?

marstamm:
Cf. https://github.com/bpmn-io/properties-panel/pull/319#issuecomment-1880708686

I tried it out and it is possible with minimum effort. As it is a one-line fix, I added this directly to this branch ([506fba7](https://github.com/bpmn-io/properties-panel/pull/319/commits/506fba73e744f0e4e9841718512d024edc613824))
Recording 2024-01-08 at 11 33 38

nikku:
Looks great to me (including text area style line-breaks) :).

vsgoulart:
@marstamm The scroll padding should be fixed now

nikku:
This works nicely for me now. 

A required additional and potentially simple fix is to move the popup editor icon to the top of the input field. You'll otherwise be forced to scroll all the way down for longer FEEL expressions to toggle the popup editor :see_no_evil:.

capture icqBXw_optimized

nikku (approved):

marstamm (approved):

PR #320, opened 2023-12-18
Housekeeping + prepare v3.14.0

nikku:
This updates dependencies, and prepares the `v3.14.0` release.


nikku:
I tested it against the latest `bpmn-js-properties-panel` trunk.

PR #321, opened 2024-01-05
feat: allow tooltip re-usability

smbea:
Related to https://github.com/camunda/improved-canvas/issues/18
Can be tested against https://github.com/camunda/improved-canvas/pull/20

### DIFF
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -1,7 +1,8 @@
 /**
  * Theming
  */
-.bio-properties-panel {
+ .bio-properties-panel,
+ .djs-parent {
   --color-grey-225-10-15: hsl(225, 10%, 15%);
   --color-grey-225-10-35: hsl(225, 10%, 35%);
   --color-grey-225-10-55: hsl(225, 10%, 55%);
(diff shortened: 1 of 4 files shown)


src/assets/properties-panel.css
```diff
@@ -1,7 +1,7 @@
 /**
  * Theming
  */
-.bio-properties-panel {
+:root {
```
nikku, line 4: Let's consider to have a shared base class instead, [i.e. `djs-parent`](https://github.com/bpmn-io/diagram-js/blob/develop/assets/diagram-js.css#L4) or something. The alternative is that we prefix our variables (which we don't usually do at bpmn.io).

Without any of these measures we risk to clash with other variables that the user defined. These names sound _very_ common to me.

nikku (changes requested):
Let's address https://github.com/bpmn-io/properties-panel/pull/321/files#r1445841787 and we're good.

src/assets/properties-panel.css
smbea: I doubt that we clash since the names are the color spefication (eg `color-grey-225-10-15`). At worst, we would override with the same color. But for safety, I added `djs-parent` (needs to have both that and `bio-properties-panel` since the latter is out of the parent)

nikku (approved):

PR #332, opened 2024-01-24
feat: remove feel input reziser

smbea:
Related to https://github.com/camunda/camunda-modeler/issues/3895

Since the input is auto-resized, FEEL popup editor can be opened to see it better and the FEEL icon is getting in the way of the resizer, i followed the issues' suggestion and removed it.

### DIFF
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -1113,7 +1113,7 @@ textarea.bio-properties-panel-input {
 }
 
 .bio-properties-panel-feel-editor-container .bio-properties-panel-input {
-  resize: vertical;
+  resize: none;
   overflow: hidden;
   overflow-y: auto;
 }


nikku (commented):
Tested it and it works. 

One minor thing you could consider: https://github.com/bpmn-io/properties-panel/pull/332/files#r1464546987

One additional thing I found is that the popup editor toggle is being pushed down by big contents, leaving folks unable to open it :see_no_evil: 

image

I believe that one could be a nice follow-up fix.

smbea:
I updated with your suggestion. It works without but I guess this is safer.

> One additional thing I found is that the popup editor toggle is being pushed down by big contents, leaving folks unable to open it 🙈

So it should stay at the top instead for visibility?

nikku (approved):

PR #333, opened 2024-01-24
fix(Entries): re-validate when validation function changes

marstamm:
related to https://github.com/camunda/camunda-modeler/issues/3070

Root Cause here: https://github.com/camunda/camunda-modeler/issues/3070#issuecomment-1908175398

### DIFF
--- a/src/components/entries/FEEL/Feel.js
+++ b/src/components/entries/FEEL/Feel.js
@@ -543,7 +543,7 @@ export default function FeelEntry(props) {
 
       setValidationError(newValidationError);
     }
-  }, [ value ]);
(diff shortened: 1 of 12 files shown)


barmac:
We shot ourselves in the foot: https://github.com/bpmn-io/properties-panel/blob/main/.eslintrc#L11

barmac (approved):

marstamm:
I will create a follow-up 👍

marstamm:
image

We'll want to fix the warnings while we do it. I created follow-up issues to tackle this in a structured manner for all libraries.

PR #341, opened 2024-01-30
fix: prevent button from taking entire height

vsgoulart:
@nikku There was a missing fix on the popup editor button from [this PR](https://github.com/bpmn-io/properties-panel/pull/337), on `form-js` it was still wrong
image

### DIFF
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -1433,10 +1433,6 @@ textarea.bio-properties-panel-input {
   display: block;
 }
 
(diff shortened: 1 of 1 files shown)


nikku (approved):

nikku:
@vsgoulart I also realize the icon is still wrong, is this something we have to fix in addition?

vsgoulart:
> @vsgoulart I also realize the icon is still wrong, is this something we have to fix in addition?

@nikku it's something we have to fix in addition, @Skaiir noticed it too

I'll create an issue

vsgoulart:
Done https://github.com/bpmn-io/form-js/issues/1026

Issue #311, opened 2023-11-29
Do not sort entries alphabetically


nikku:
### Is your feature request related to a problem? Please describe.

With the properties panel rewrite we decide to make sorting of entries alphabetically the default. Since then we've unbuilt in the BPMN properties panel the sorting for most parts.

We did it because sorting alphabetically for anything where ordering (in XML) has an execution semantic is dangerous (https://github.com/camunda/camunda-modeler/issues/2932#issuecomment-1804223456). If at all, then users should have the chance to sort things manually.

Unfortunately the fact that sorting is the default continues to bite us, cf. https://github.com/camunda/camunda-modeler/issues/3978. I therefore propose to disable sorting per default, and instead understand that the properties panel operates in a domain where deterministic, user chosen sorting is the standard.

### Describe the solution you'd like

* Disable alphabetically sorting (make `shouldSort=false` per default)

### Describe alternatives you've considered

* Get rid of sorting by some label all together

### Additional context

Raised last (again) in the context of C7: https://github.com/camunda/camunda-modeler/issues/3978.

nikku:
This is not going to happen. Moving back to `backlog`.

PR #360, opened 2024-05-28
feat: scroll focused element into view

nikku:
This ensures that a newly created (and focused) entry is in view:

capture Giz5NV_optimized

Related to #311


barmac:
I'm testing this with `npx @bpmn-io/sr -l bpmn-io/properties-panel#scroll-to-focused-element bpmn-io/bpmn-js-properties-panel`

barmac (approved):
Works like a charm

PR #362, opened 2024-05-28
Enhance ListGroup and Collapsible components to support translation and add tests for translation

### LINKED ISSUE #358: Add missing translation capabilities to ListGroup
### Is your feature request related to a problem? Please describe.

ListGroup has a create, delete and toggle button which all lack translations. Tooltips for these buttons as well as the indicator for how many items the listGroup contains, also lack translations. 

### Describe the solution you'd like

It should be possible to create translations for these buttons and their tooltips. We could achieve this by passing the translate function as a prop.

### Additional context

grafik

grafik

grafik

grafik

ulli-hoeness:
Closes #358 

@barmac Sorry for opening a new pull request. I was signed in with the wrong user and was unable to sign the CLA.

### DIFF
--- a/src/components/ListGroup.js
+++ b/src/components/ListGroup.js
@@ -26,6 +26,7 @@ import {
 import { PropertiesPanelContext } from '../context';
 
 import { useStickyIntersectionObserver } from '../hooks';
+
 import translateFallback from './util/translateFallback';
 
 const noop = () => {};
--- a/src/components/ListGroup.js
+++ b/src/components/ListGroup.js
@@ -26,6 +26,7 @@ import {
 import { PropertiesPanelContext } from '../context';
 
 import { useStickyIntersectionObserver } from '../hooks';
+import translateFallback from './util/translateFallback';
 
 const noop = () => {};
 
@@ -40,7 +41,7 @@ export default function ListGroup(props) {
     items,
     label,
     shouldOpen = true,
-    translate
+    translate = translateFallback
   } = props;
 
   useEffect(() => {
@@ -168,14 +169,14 @@ export default function ListGroup(props) {
             ? (
               <button
                 type="button"
-                title={ translate ? translate('Create new list item') : 'Create new list item' }
+                title={ translate('Create new list item') }
                 class="bio-properties-panel-group-header-button bio-properties-panel-add-entry"
                 onClick={ handleAddClick }
               >
                 <CreateIcon />
                 {
                   !hasItems ? (
-                    <span class="bio-properties-panel-add-entry-label">{ translate ? translate('Create') : 'Create' }</span>
+                    <span class="bio-properties-panel-add-entry-label">{ translate('Create') }</span>
                   )
                     : null
                 }
@@ -187,11 +188,7 @@ export default function ListGroup(props) {
           hasItems
             ? (
               <div
-                title={
-                  translate
-                    ? translate(`List contains {numOfItems} item${items.length != 1 ? 's' : ''}`, { numOfItems: items.length })
-                    : `List contains ${items.length} item${items.length != 1 ? 's' : ''}`
-                }
+                title={ translate(`List contains {numOfItems} item${items.length != 1 ? 's' : ''}`, { numOfItems: items.length }) }
                 class={
                   classnames(
                     'bio-properties-panel-list-badge',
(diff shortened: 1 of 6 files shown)


nikku (changes requested):
@ulli-hoeness Please update on latest `main`, there is a couple of changes that your change conflict with (removed sorting).

src/components/ListGroup.js
```diff
@@ -185,7 +186,7 @@ export default function ListGroup(props) {
           hasItems
             ? (
               <div
-                title={ `List contains ${items.length} item${items.length != 1 ? 's' : ''}` }
+                title={ translate(`List contains ${items.length} item${items.length != 1 ? 's' : ''}`) }
```
nikku, line 189: Proposal: Feed number of items in as a context argument:

```suggestion
                title={ translate(`List contains {items} item${items.length != 1 ? 's' : ''}`, { items: items.length }) }
```

You'll otherwise need to provide an infinite list of translation strings.

ulli-hoeness: Thank you! I changed this line.

ulli-hoeness:
> @ulli-hoeness Please update on latest `main`, there is a couple of changes that your change conflict with (removed sorting).

I merged main into my branch. I see the merge locally but i do not see it in the commits in this pull request although i have pushed everything. Working with github is new to me so I apologize if the changes are not as requested. Please have a look and if my branch is still behind, I will fix it.

nikku:
@ulli-hoeness You'd want to rebase your branch on top of main for a clear history:

```
git checkout main
git pull
git checkout your-feature-branch
git rebase main
```

You'll be able to see merge conflicts (if they appear), fix them and have a clean history.

nikku:
@ulli-hoeness Could you follow-up according to my latest feedback (https://github.com/bpmn-io/properties-panel/pull/362#issuecomment-2139102744)? We don't do merges in this library, as we emphasize a clean history.

ulli-hoeness:
> @ulli-hoeness You'd want to rebase your branch on top of main for a clear history:
> 
> ```
> git checkout main
> git pull
> git checkout your-feature-branch
> git rebase main
> ```
> 
> You'll be able to see merge conflicts (if they appear), fix them and have a clean history.

@nikku Thank you! I apologize for the delayed response. Unfortunately I haven't had the time to deal with this yet. Will rebase main onto my branch within the next few days. Will update you when it is ready for review.

marstamm (changes requested):
I have rebased the branch for you. Please check the test failures:

The current implementation breaks if you use placeholders, as they are not replaced by the fallback functions, e.g.
```
translate(`List contains {numOfItems}`, { numOfItems: 15 })
// Results in "List contains {numOfItems}"
```

We want to ensure it still works without providing a translation provider.

nikku:
Cf. [base translate implementation](https://github.com/bpmn-io/diagram-js/blob/develop/lib/i18n/translate/translate.js).

ulli-hoeness:
Thank you! Will correct the placeholder behaviour today.

ulli-hoeness:
> I have rebased the branch for you. Please check the test failures:
> 
> The current implementation breaks if you use placeholders, as they are not replaced by the fallback functions, e.g.
> 
> ```
> translate(`List contains {numOfItems}`, { numOfItems: 15 })
> // Results in "List contains {numOfItems}"
> ```
> 
> We want to ensure it still works without providing a translation provider.


@marstamm  Thank you for the Rebase and pointing out the bug in the code. Please have a look if the conditional use of the translate function in the newest commit is okay. You pointed out correctly that the previously used fallback function of (string) => string does not work as desired for string which use raplace values. Since no other component contained string which use replace values, I did leave the fallvack function for translate as it was. Only in ListGroup we check if translate exists.

marstamm:
Thank you for the follow-up! While this works for the current state, we introduce 2 patterns to deal with missing translation provider. Either we have a fallback function or use the ternary. We should aim to have a consistent pattern for translations.

Having a fallback translation provider as a utility function (cf. [Nikku's comment](https://github.com/bpmn-io/properties-panel/pull/362#issuecomment-2165138356)) would be a more elegant solution. It would ensure that we can use `translate` everywhere and is a pattern we use in other libraries as well.

You can create a shared utility in https://github.com/bpmn-io/properties-panel/tree/main/src/components/util and use it as a fallback in both locations

ulli-hoeness:
As requested, I have used the linked code for the translate fallback and have placed it in src/components/util. This fallback is now used in case no translation module is present

marstamm:
Thank you for the follow-up! Looks good to me now, I'll go ahead and release the changes. Thank you for your contribution 🏆

marstamm:
Released as [v3.21.0](https://github.com/bpmn-io/properties-panel/releases/tag/v3.21.0)

PR #374, opened 2024-07-17
fix: text area resizes properly every time it becomes visible

### LINKED ISSUE #314: TextAreaEntry with autoResize not working after toggle section and change selection
### Describe the Bug

When toggled section and then changed the selection to another shape, I found the `TextAreaEntry` with autoResize degrades to only 1 row.

I made a minimal reproducible example in this [repository](https://github.com/ptyin/properties-panel-poc).

### Steps to Reproduce

The following steps can be reproduced on this demo [site](https://ptyin.github.io/properties-panel-poc/).

1. `TextArea` with autoResize
  
image

2. toggle the section
  
image

3. change the selection and toggle the section again
image

### Expected Behavior

I believe it should be re-autoResize even after toggling and selection.

### Environment
- Library version: [e.g. 3.13.0]

---

Related to https://github.com/camunda/camunda-modeler/issues/4419

jarekdanielak:
Closes #314

### Problem

Text area with `autoResize` property only resizes on initial render and on input. When it's initially rendered inside of closed group or collapsable, it won't properly resize after opening the container.

https://github.com/user-attachments/assets/47a1a43a-078d-4309-a617-4bcd52b4574e

### Solution

Watch for change of element's visibility and trigger resize when it becomes visible. Since when toggling the container, the text area's style doesn't change directly, I use `ResizeObserver` to detect change of element's size.

https://github.com/user-attachments/assets/c18f790e-8aa2-44e3-a67c-ae902238661c

### Checklist

* [X] **Brief textual description** of the changes present
* [X] **Visual demo** attached
* [X] Related issue linked via `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`

### DIFF
--- a/src/hooks/useElementVisible.js
+++ b/src/hooks/useElementVisible.js
@@ -10,8 +10,15 @@ export function useElementVisible(element) {
     const resizeObserver = new ResizeObserver(([ entry ]) => {
 
       // Prevents ResizeObserver from going into infinite loop.
-      requestAnimationFrame(() => setVisible(!!entry.contentRect.height));
+      requestAnimationFrame(() => {
+
+        const newVisible = !!entry.contentRect.height;
+        if (newVisible !== visible) {
+          setVisible(newVisible);
+        }
+      });
     });
+
     resizeObserver.observe(element);
 
     return () => resizeObserver.disconnect();
--- a/src/hooks/useElementVisible.js
+++ b/src/hooks/useElementVisible.js
@@ -2,27 +2,22 @@ import { useLayoutEffect, useState } from 'preact/hooks';
 
 export function useElementVisible(element) {
 
-  const [ visible, setVisible ] = useState(false);
+  const [ visible, setVisible ] = useState(!!element && !!element.clientHeight);
 
   useLayoutEffect(() => {
     if (!element) return;
 
     const resizeObserver = new ResizeObserver(([ entry ]) => {
-
-      // Prevents ResizeObserver from going into infinite loop.
-      requestAnimationFrame(() => {
-
-        const newVisible = !!entry.contentRect.height;
-        if (newVisible !== visible) {
-          setVisible(newVisible);
-        }
-      });
+      const newVisible = !!entry.contentRect.height;
+      if (newVisible !== visible) {
+        setVisible(newVisible);
+      }
     });
 
     resizeObserver.observe(element);
 
     return () => resizeObserver.disconnect();
-  }, [ element ]);
+  }, [ element, visible ]);
 
   return visible;
 }
--- a/src/hooks/useElementVisible.js
+++ b/src/hooks/useElementVisible.js
@@ -8,10 +8,12 @@ export function useElementVisible(element) {
     if (!element) return;
 
     const resizeObserver = new ResizeObserver(([ entry ]) => {
-      const newVisible = !!entry.contentRect.height;
-      if (newVisible !== visible) {
-        setVisible(newVisible);
-      }
+      requestAnimationFrame(() => {
+        const newVisible = !!entry.contentRect.height;
+        if (newVisible !== visible) {
+          setVisible(newVisible);
+        }
+      });
     });
 
     resizeObserver.observe(element);
--- a/src/hooks/useElementVisible.js
+++ b/src/hooks/useElementVisible.js
@@ -1,0 +1,21 @@
+import { useLayoutEffect, useState } from 'preact/hooks';
+
+export function useElementVisible(element) {
+
+  const [ visible, setVisible ] = useState(false);
+
+  useLayoutEffect(() => {
+    if (!element) return;
+
+    const resizeObserver = new ResizeObserver(([ entry ]) => {
+
+      // Prevents ResizeObserver from going into infinite loop.
+      requestAnimationFrame(() => setVisible(!!entry.contentRect.height));
+    });
+    resizeObserver.observe(element);
+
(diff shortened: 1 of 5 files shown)


test/spec/components/TextArea.spec.js
```diff
@@ -458,7 +460,9 @@ GEHTS
       const initialHeight = input.clientHeight;
 
       // then
-      expect(initialHeight).to.be.greaterThan(60);
+      waitFor(() => {
```
barmac, line 463: Note that the test won't wait for the result as it exits right after the `waitFor` call. So the test will pass even though the callback may fail. We should either `await` the result or return the Promise from the test.

```suggestion
      return waitFor(() => {
```

test/spec/components/TextArea.spec.js
```diff
@@ -517,6 +521,37 @@ GEHTS
...
+
+      // then
+      expect(hiddenHeight).to.be.eq(0);
+
+      // when
+      container.style.display = 'block';
+      const visibleHeight = input.clientHeight;
+
+      // then
+      waitFor(() => {
```
barmac, line 550: Same as https://github.com/bpmn-io/properties-panel/pull/374/files#r1681148118

test/spec/components/TextArea.spec.js
```diff
@@ -517,6 +521,37 @@ GEHTS
...
+        expect(visibleHeight).to.be.greaterThan(initialHeight);
```
barmac, line 551: I think we should make sure that the height as more or less equal to 4 lines as it is not good enough to have a single line of text height.

src/hooks/useElementVisible.js
```diff
@@ -0,0 +1,21 @@
...
+
+  const [ visible, setVisible ] = useState(false);
+
+  useLayoutEffect(() => {
+    if (!element) return;
+
+    const resizeObserver = new ResizeObserver(([ entry ]) => {
+
+      // Prevents ResizeObserver from going into infinite loop.
+      requestAnimationFrame(() => setVisible(!!entry.contentRect.height));
```
barmac, line 13: If I understand it correctly, the resize observer may still request multiple animation frames. Perhaps we could cache the last known value and change `visible` value only on change?

src/hooks/useElementVisible.js
```diff
@@ -0,0 +1,21 @@
+import { useLayoutEffect, useState } from 'preact/hooks';
+
+export function useElementVisible(element) {
```
barmac, line 3: Can we have a test for the hook specifically?

test/spec/components/TextArea.spec.js
jarekdanielak: It seems cumbersome to calculate the exact height it should be. but checking for twice the initial size should be enough.

test/spec/components/TextArea.spec.js
```diff
@@ -435,33 +437,6 @@ describe('<TextArea>', function() {
 
   describe('auto resize', function() {
 
-    it('should resize initially', function() {
```
jarekdanielak, line 438: Removed this test since there is a valid use case when text area doesn't need to resize on initial render. When it's initially rendered inside of a closed group/collapsable, it will resize when it becomes visible.

barmac:
I pushed two commits on top:
* https://github.com/bpmn-io/properties-panel/pull/374/commits/afac8fdd5662610fd28689dd511a56c3d37c7b32 to make the tests execute assertions correctly 
* https://github.com/bpmn-io/properties-panel/pull/374/commits/bdf8f13a5453db0cf4b61ed4b2bb6e51f55dae02 to simplify the hook

jarekdanielak:
@barmac ready for final approval. 🎉

test/spec/hooks/useElementVisible.spec.js
```diff
@@ -0,0 +1,118 @@
...
+    container.appendChild(element);
+
+    // when
+    const { rerender, result } = renderHook(() => useElementVisible(element));
+
+    // then
+    await waitFor(() => {
+      rerender();
+      expect(result.current).to.be.true;
+    });
```
barmac, line 31: [not blocking] Since this and the following two cases test a static result, we could even check the immediate result of the hook. This way we can ensure that the value is correct from the beginning, and not changed in the meantime (eventually correct).

barmac: So
```suggestion
expect(result.current).to.be.true
```

(async can be removed as well)

test/spec/hooks/useElementVisible.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+    // then
+    await waitFor(() => {
+      rerender();
+      expect(result.current).to.be.false;
+    });
+  });
+
+
+  it('should return true when element becomes visible', async function() {
```
barmac, line 67: 🚀

test/spec/hooks/useElementVisible.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+    // then
+    await waitFor(() => {
+      rerender();
+      expect(result.current).to.be.true;
+    });
+  });
+
+
+  it('should return true when element becomes hidden', async function() {
```
barmac, line 94: Typo:
```suggestion
  it('should return false when element becomes hidden', async function() {
```

barmac:
I left two comments to fix before merging. If the CI is green after the test changes, I think you can just squash it, and I am looking forward to having this in the Modeler :)

PR #390, opened 2024-11-26
fix: focus offset added for checkbox and buttons

abdul99ahad:
Related to https://github.com/camunda/camunda-modeler/issues/4632

### Proposed Changes

 
Fixed the focus outline offset

Checkbox: 
image

Button: 
image

### Checklist


src/assets/properties-panel.css
```diff
@@ -299,6 +301,10 @@
   background: none;
 }
 
+.bio-properties-panel-group-header-button:focus {
+  outline-offset: 2px;
```
philippfromme, line 305: Outline still black for me:

brave_gxcOzPHfJA

Suggestion:

```suggestion
  outline-offset: 2px;
  outline-color: var(--focus-outline-color);
```

abdul99ahad: That's strange, it's already showing blue for me. Are you linking the library to `bpmn-js-properties-panel` to verify the changes?

Even before my updates, the outline was already blue; I only added the offset. Could you double-check? I want to avoid adding redundant properties.

philippfromme: Just checked again and it's still black. Where would the blue color come from?

abdul99ahad: It was coming from `-webkit-focus-ring-color` (default focus color). I have updated the code and added the outline color property.

src/assets/properties-panel.css
```diff
@@ -17,6 +17,7 @@
   --color-blue-205-100-35: hsl(205, 100%, 35%);
   --color-blue-205-100-45: hsl(205, 100%, 45%);
   --color-blue-205-100-50: hsl(205, 100%, 50%);
+  --color-blue-205-100-40: hsl(205, 100%, 40%);
```
philippfromme, line 20: Nitpick: Let's maintain order by luminance.

src/assets/properties-panel.css
```diff
@@ -537,8 +544,9 @@ select.bio-properties-panel-input {
 }
 
 .bio-properties-panel-input[type="checkbox"]:focus {
-  outline: 2px solid var(--input-focus-border-color);
-  outline-offset: 0;
+  outline: auto;
```
philippfromme, line 547: Why `auto`? It let's the user agent to render a custom outline style. Is this what we want?

philippfromme: Then I'd expect it to be consistent. If we already set the color we should also set the style. As you said, one can always override.

abdul99ahad: I agree. We should maintain consistent styling across all child libraries instead of relying entirely on integrators. While we can allow them the flexibility to override styles, our defaults should not depend on their configurations.

abdul99ahad: We've already specified the color in the same class. Using the auto property ensures an default outline is provided, overriding the none outline for inputs.

```javascript
.bio-properties-panel-input:focus,
.bio-properties-panel-input:focus-within {
  outline: none;
```

src/assets/properties-panel.css
abdul99ahad: Done ✅

nikku (commented):
Looks good to me, looking to @philippfromme for final check.

philippfromme:
Are we okay with this outline still being black?

image

nikku:
Yes, because outline color is setup by the application.

nikku:
@abdul99ahad ready to be released and integrated :heavy_check_mark:

PR #397, opened 2025-01-31
Sticky list headers

nikita-kart:
### Proposed Changes

 

Related to https://github.com/bpmn-io/bpmn-js-properties-panel/issues/1103

These changes allow list headers to become sticky to the properties panel when scrolled past. This is particularly useful when managing long lists of values, as it eliminates the need to constantly scroll back up to add new elements to a list or map. It is especially convenient for maps where each entry contains multiple fields making scrolling distance quite long.

Additionally, there is the ability to set nesting levels for headers, which is useful when working with multiple nested lists or maps. This value is used to set z-index and top margin for the sticky position and to properly detect stickiness with `useStickyIntersectionObserver` hook.

A value of 1 means that the list is at the first nesting level within its group section. If this property is not set, the header will behave as before.

#### How it looks like

Sticky list headers

### Checklist

### DIFF
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -827,7 +827,6 @@ textarea.bio-properties-panel-input {
 }
 
 .bio-properties-panel-list-entry-header.position-sticky {
-  position: -webkit-sticky;  /* for safari */
   position: sticky;
   top: 0;
   z-index: 5;
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -826,6 +826,18 @@ textarea.bio-properties-panel-input {
   height: 32px;
 }
 
+.bio-properties-panel-list-entry-header.position-sticky {
+  position: -webkit-sticky;  /* for safari */
+  position: sticky;
+  top: 0;
+  z-index: 5;
+}
+
+.bio-properties-panel-list-entry-header.sticky {
+  background-color: var(--color-white);
+  border-bottom: 1px solid var(--sticky-group-bottom-border-color);
+}
+
 /* Nested list dot */
 .bio-properties-panel-list-entry::before {
   content: "";
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -819,22 +819,18 @@ textarea.bio-properties-panel-input {
 }
 
 .bio-properties-panel-list-entry-header {
-  position: relative;
+  position: sticky;
   overflow: hidden;
   display: flex;
   justify-content: space-between;
   height: 32px;
 }
 
-.bio-properties-panel-list-entry-header.position-sticky {
-  position: sticky;
-  top: 0;
-  z-index: 5;
-}
-
 .bio-properties-panel-list-entry-header.sticky {
   background-color: var(--color-white);
   border-bottom: 1px solid var(--sticky-group-bottom-border-color);
+  top: 32px;
+  z-index: 10;
 }
 
 /* Nested list dot */
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -830,7 +830,7 @@ textarea.bio-properties-panel-input {
   background-color: var(--color-white);
   border-bottom: 1px solid var(--sticky-group-bottom-border-color);
   top: 32px;
-  z-index: 10;
+  z-index: 9;
 }
 
 /* Nested list dot */
--- a/src/components/entries/List.js
+++ b/src/components/entries/List.js
@@ -1,6 +1,7 @@
 import {
   useEffect,
-  useState
+  useState,
+  useRef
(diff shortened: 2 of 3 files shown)


jarekdanielak:
Great contribution! 👏

I think we could consider making stickiness the default behavior for the list headers, like it is for the group headers. What do you think @barmac?

Let's have our designer take a look too. cc. @lmbateman (Task > Inputs > Assignment type > Map > Map entries)

I prepared a test branch in `bpmn-js-properties-panel` where it's enabled for map entries:

```
npx @bpmn-io/sr -c="npm run start:platform" -l="nikita-kart/properties-panel#sticky-list-headers" bpmn-io/bpmn-js-properties-panel#header-nesting
```

I'll continue with a proper code review when we decide if we want to make it the new default.

src/assets/properties-panel.css
```diff
@@ -826,6 +826,18 @@ textarea.bio-properties-panel-input {
   height: 32px;
 }
 
+.bio-properties-panel-list-entry-header.position-sticky {
+  position: -webkit-sticky;  /* for safari */
```
jarekdanielak, line 830: `position: sticky` is now [widely available for all major browsers](https://caniuse.com/?search=sticky). I think we can skip it.

lmbateman:
Very nice! I don't really have any comments; it transitions smoothly from header to header and back to 0. I did notice that I can't seem to get subheaders under Execution Listeners, even though the original GIF shows them...

image

barmac:
> I think we could consider making stickiness the default behavior for the list headers, like it is for the group headers. What do you think @barmac?

I agree. It works great for the groups, so there's no reason to not make it the default for lists too.

src/assets/properties-panel.css
nikita-kart: Thanks for pointing that out, I removed -webkit-sticky

nikita-kart:
Thanks for feedback!

> let's try adding a left border and see how it looks

I tried that, here's the result:
image

Without changing left margin of the header text it looks kinda strange to me, so I also tried to set additional margin when it becomes sticky:


Looks better but there is this "jump" now when the header moves to the right a bit

> I did notice that I can't seem to get subheaders under Execution Listeners. Will this be extended to all multi-level sections?

I need a little clarification here. Are you talking about every foldable element in panel and not just lists and maps? I did not tried to that, but I might look into it when I'll have a moment

lmbateman:
> Looks better but there is this "jump" now when the header moves to the right a bit

Thanks for trying this out! I'm guessing it would not be easy to move the left border to the left a few pixels to prevent the jump. If that's correct, then we should probably stay with the version without the left border.

barmac:
> It misses the edge case where the nested header is briefly hidden behind the main header. However, this isn’t noticeable during normal use of the properties panel, so I don’t think it’s worth adding extra logic to the Intersection Observer just to handle it.

Could you please share a screencast?

barmac:
Is this what you meant?

https://github.com/user-attachments/assets/2ffedf9d-8de9-4609-ad11-0a338c36f9cb

barmac:
This problem can be solved if we lower the z-index of the nested header.

https://github.com/user-attachments/assets/27135abc-e73c-4f3c-be8f-7976c7ab4a9e

nikku:
I tried this out locally, via `npm run start:platform` in the bpmn-js-properties-panel project.

Overall it works correctly, and is a simple solution. However across list items and now map headers being sticky I miss the actual elements to be sticky, too. The fact that I am in the inputs section is now preserved, the fact that I'm within an input's map section is preserved. What I don't know is _what input I'm currently manipulating_.

capture Shgh6U_optimized

barmac:
Indeed, we'd need to implement sticky [Collapsible](https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/Collapsible.js) headers to achieve what you propose @nikku. That should close the story.

jarekdanielak:
@barmac [this](https://github.com/bpmn-io/properties-panel/pull/397#issuecomment-2966693852) is not what I meant. Fixed with lowering `z-index` on the nested header (https://github.com/bpmn-io/properties-panel/pull/397/commits/8fe5794051c90c9eb605a145eea1352a20555d09).

@nikku it's indeed a nice improvement, but I couldn't get the stickiness to work on the collapsible. We would have to account for a couple of different nesting levels, other sticky headers of the same level open above and indent indicators which in our implementation are not part of the actual header (see image below).

image

Again, I don't see this as a good investment, especially for C7, so I suggest we ship this as it is or park it for some contributor to fully solve.

barmac (approved):
It's already some improvement. We can leave the collapsible part out of the PR.

jarekdanielak:
@nikita-kart thanks for the contribution! 🚀

PR #398, opened 2025-02-13
fix: remove the focus border within feel popups

Skaiir:
Related to https://github.com/camunda/web-modeler/issues/12976

### Proposed Changes

This change remove the focus outline present inside of the popouts. Note this issue was popping up due to a variable naming conflict with [react-pdf](https://www.npmjs.com/package/react-pdf). I don't think it's an absolute necessity to change the variable, but something to consider. 

See the full RCA [here](https://github.com/camunda/web-modeler/issues/12976#issuecomment-2656485410).

#### Image:
image

#### Replicating the issue:
Easiest way I can think: just set `--input-focus-border-color` to something globally somewhere where you can access the popup, put focus within it.

### DIFF
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -1502,6 +1502,10 @@ textarea.bio-properties-panel-input {
   padding-left: 4px;
 }
 
+.bio-properties-panel-feel-popup .bio-properties-panel-input {
+  border: none;
+}
+
 .bio-properties-panel-feel-popup .cm-gutters {
   background-color: var(--feel-popup-gutters-background-color);
   border: none;


nikku:
@Skaiir I'm not sure we want to go this direction, as it introduces CSS complexity. I regard that as technical dept. From what I see we have an _isolation problem, that can be solved in two ways:

* Ask `react-pdf` to properly prefix their variables.
* Ask `@bpmn-io/properties-panel` to properly prefix their variables

To be future prove I propose that our libraries _should_ properly be prefixed, `bio`, for example. If you consider UI frameworks or larger applications, they do the same. Carbon for example prefixes every variable with `cds` ([ref](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/sass.md#prefixes)).

Skaiir:
@nikku While I agree on the prefixing, I still think we should prevent the border actively. If there is every some outlining, it's not going to be driven by a border because that shifts the text around. Either way the styles, as they stand, are not correct, relying only on a prefix would make them not correct but hidden because the variable happens to not be defined, which doesn't feel like a good state to me either.

nikku:
> While I agree on the prefixing, I still think we should prevent the border actively. 

Ok, your take is: Independent from the clash the border makes no sense? Or is there situations where it would make sense?

Skaiir:
> Independent from the clash the border makes no sense? 

Yup, exactly. If there was a situation that called for some kind of a highlight, we'd define it differently. But as it stands, it's an undesired behavior, and we only coincidentally spotted it because of the clash.

nikku:
I double-checked this with the other inputs we have. In fact it is the other way around:

* Focus _is_ indicated through a border, and outline is usually disabled.

  image

----

The popover simply has no valid focus style at the moment, while other FEEL inputs have:

_FEEL input, indicating focus:_

screenshot osAzTn

_Popover, not indicating focus:_

screenshot 5ufJtc

I'll ask the team for further input on this.

Skaiir:
This is where the styling actually comes from, the smaller input div is replicated within the pop-out, to inherit some of its characteristics. But for example, the gray non-focused border unlike the highlight border, is not rendered at the moment, so everything gets shifted 1pixel inward when the highlight border is drawn.

So if we do want borders here, we'd want to consistently apply it whether it's focused or not (gray/transparent unfocused).  I would find it odd from a design perspective because this pop-out is distinct enough that it doesn't really need outlining from its surroundings, but this is something that can be discussed with, well, designers.

Skaiir:
Btw just because I haven't mentioned this yet I believe but web modeler core confirmed with me when I raised this that indeed they don't expect a border here. Of course it's just one of the stakeholders, but just for transparency.

barmac:
I'd expect the focus to be indicated via outline. Not sure why we have implemented this via border in the first place.

barmac:
As for whether we should or should not indicate focus in the popup, I think it's fair to follow the existing pattern and always display the focus (https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html).

lmbateman:
As Maciej said, best practice is to use outline rather than border.

I did a review of several applications and of WCAG, and learned:
- The trend seems to be to use only the cursor to indicate focus in a text editor (VS Code, Eclipse, Sublime Text, Monaco).
- https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html offers two examples, the first of which is, "When text fields receive focus, a vertical bar is displayed in the field, indicating that the user can insert text, OR all of the text is highlighted, indicating that the user can type over the text."

So it sounds like the blinking cursor is an acceptable alternative to the outline.

src/assets/properties-panel.css
```diff
@@ -1502,6 +1502,11 @@ textarea.bio-properties-panel-input {
   padding-left: 4px;
 }
 
+.bio-properties-panel-feel-popup .bio-properties-panel-input:focus,
+.bio-properties-panel-feel-popup .bio-properties-panel-input:focus-within {
+  border: none;
```
nikku, line 1507:
```suggestion
.bio-properties-panel-feel-popup .bio-properties-panel-input,
.bio-properties-panel-feel-popup .bio-properties-panel-input {
  border: none;
```

Skaiir:
> Why only remove the focus border? Don't we want to simply fix the border color, or remove the border entirely?

As far as I know, the border is not expected, does not show up within the library and the old state did not include it. Web modeler agreed with me on this. Because this wasn't present before, and only showed up accidentally over time due to this variable clash, my take would be that, not removing it would be considered a "change", and as per the [design comments](https://github.com/bpmn-io/properties-panel/pull/398#issuecomment-2663349455) on this thread, a change which is not necessary.

Skaiir:
* I misread, changing the message.

> [...] remove the border entirely?

Totally agree, I just need to check if the CSS specificity is enough.

Skaiir:
Generalization is tested, works with the border variable set. Pushed the changes.

nikku:
Good stuff :clap:. I'm going ahead and release this.

nikku:
Released as [v3.26.1](https://github.com/bpmn-io/properties-panel/releases/tag/v3.26.1)

PR #399, opened 2025-02-20
fix: make tooltip persist when trying to copy from it + refactor

jarekdanielak:
### Proposed Changes

I reworked the Tooltip component to rely on the React event handlers instead of hooking into document events and running hundreds of checks for every `mousemove`, `wheel` and `focusout` event.

In the process, I fixed the bug that made the tooltip disappear when you click on it to try to copy something after opening it with a click on the header, not hovering. (cf. https://github.com/camunda/camunda-modeler/pull/4841#issuecomment-2671382075)

The removed test `should not render if trigger element no longer hovered - scroll` doesn't seem to make sense. Scrolling shouldn't hide the tooltip itself, only if it actually moves the cursor out of it. If it does, it fires `mouseleave` event anyway.

Interesting finding: potentially we could rely on [sourceCapabilities](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/sourceCapabilities) property of FocusEvent to tell if it was triggered by mouse or keyboard, but it's still experimental.

Looks a'ight.

https://github.com/user-attachments/assets/6b376427-71a5-449d-9fd2-df32ced3238a

### Try it 

```
npx @bpmn-io/sr -l="bpmn-io/properties-panel#tooltip-refactor" bpmn-io/bpmn-js-properties-panel
```

Related to https://github.com/camunda/camunda-modeler/issues/4834

### Checklist

### DIFF
--- a/src/components/entries/Tooltip.js
+++ b/src/components/entries/Tooltip.js
@@ -1,6 +1,5 @@
 import {
   useRef,
-  useEffect,
   useState
 } from 'preact/hooks';
 
@@ -27,7 +26,7 @@ export default function TooltipWrapper(props) {
     return props.children;
   }
 
-  return <Tooltip { ...props } value={ value } forId={ prefixId(forId) } />;
+  return <Tooltip { ...props } value={ value } forId={ `bio-properties-panel-${ forId }` } />;
 }
 
 function Tooltip(props) {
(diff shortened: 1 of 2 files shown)


jarekdanielak:
You can still do some funky things with the tooltip like spawn multiple using keyboard focus and mouse at the same time or put them in weird places when resizing the window, but that all carries from the old implementation.

nikku (approved):
I think this is a huge implification of the existing code, less in complexity, and hence less fragile :tada: 

Tested manually, and works. :heavy_check_mark: 

---

One quirk I found (just a bit unexpected) is that multiple tooltips can still be open across mixed mouse and keyboard usage - arguably an edge case we should ignore:

capture uX7Ct7_optimized

barmac:
The implementation looks clean now, and I cannot reproduce the bug previously reported. The selection works fine 👍 

> One quirk I found (just a bit unexpected) is that multiple tooltips can still be open across mixed mouse and keyboard usage - arguably an edge case we should ignore:

I agree that this isn't something that we should be concerned about.

src/components/entries/Tooltip.js
```diff
@@ -27,7 +26,7 @@ export default function TooltipWrapper(props) {
     return props.children;
   }
 
-  return <Tooltip { ...props } value={ value } forId={ prefixId(forId) } />;
+  return <Tooltip { ...props } value={ value } forId={ `bio-properties-panel-${ forId }` } />;
```
barmac, line 29: This one is arguable. I'd rather keep the helper as it separates the prefixing from the rendering code.

src/components/entries/Tooltip.js
```diff
@@ -39,82 +38,59 @@ function Tooltip(props) {
...
-    );
-  };
+  const handleMouseLeave = ({ relatedTarget }) => {
 
-  useEffect(() => {
-    const { current } = wrapperRef;
-
-    if (!current) {
+    // Don't hide the tooltip when moving mouse between the wrapper and the tooltip.
+    if (relatedTarget === wrapperRef.current || relatedTarget === tooltipRef.current || relatedTarget?.parentElement === tooltipRef.current) {
```
barmac, line 70: I'm so happy that we adopted optional chaining 🚀

jarekdanielak:
Created an issue for multiple tooltips bug: https://github.com/bpmn-io/properties-panel/issues/400

barmac (approved):

PR #401, opened 2025-02-24
trim whitespaces when user leaves the field

abdul99ahad:
Closes #309

### Proposed Changes
Extra whitespaces at the end are trimmed in both text area and text field
 
Feb-24-2025 16-10-59

You can test this works via 

```
npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel -l bpmn-io/properties-panel#trim-whitespaces
```
### Checklist

To ensure you provided everything we need to look at your PR:

* [x] **Brief textual description** of the changes present
* [x] **Visual demo** attached
* [x] **Steps to try out** present, i.e. [using the `@bpmn-io/sr` tool](https://github.com/bpmn-io/sr)
* [x] Related issue linked via `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`

Issue #309, opened 2023-11-03
Auto-trim leading whitespaces when pasting into property fields


christian-konrad:
### Is your feature request related to a problem? Please describe.

When you copy things like URLs and paste them into the properties panel, it can happen that you accidently copy leading white spaces. They will be copied into the property field, and you might not spot that easily. This can lead to problems such as regex validation failing (in the URL example), and the user does not understand why.

![image](https://github.com/bpmn-io/properties-panel/assets/56470362/e5df9c59-85a9-4e03-98c1-c489ba9097cf)

https://github.com/bpmn-io/properties-panel/assets/56470362/2900abde-34fa-4039-91c8-2be9a9fd7b74

### Describe the solution you'd like

Auto-trim leading whitespaces when pasting into property fields.

### Describe alternatives you've considered

We need to confirm that there is no valid use case of pasting leading white spaces that we will break with such a feature.

### Additional context

Related to [SUPPORT-25217](https://jira.camunda.com/browse/SUPPORT-25217?focusedCommentId=390503&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-390503)

nikku:
This may be a quick fix for the bigger issue reported once in a while by our users: https://github.com/camunda/camunda-modeler/issues/2385.

Often times (in technical configuration) whitespace is bad, and more general trimming is what we considered in the past, but did not follow up on.

sbuettner:
Any chances we can pick this up? We had another support case where a customer created an invalid Connector config by having a whitespace character in their properties which lead to a long running incident process.

barmac:
We can take this to the next iteration planning given that it's such a pain for the users. Can you please link the support case in the top comment? We could then label the issue as `channel:support`.

sbuettner:
@barmac I dont have the permissions to do so. Here is the support ticket: https://jira.camunda.com/browse/SUPPORT-25217?focusedCommentId=390503&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-390503

barmac:
Thanks, I updated the issue accordingly.

nikku:
Follow-up to properly hook this up with debounced updates + implement properly across normal and FEEL text fields: https://github.com/camunda/web-modeler/issues/13599.

PR #405, opened 2025-03-17
Trimming handles input update and refactoring (code + tests)

### LINKED ISSUE #404: Trim whitespace on blur does not account for debounced updates
### Describe the Bug

After we implemented https://github.com/bpmn-io/properties-panel/issues/309 we see issues reported by users that, when switching elements, updates are applied to the newly selected element, like a super-powered-unintended-copy-between elements.

### Steps to Reproduce

Cf. https://github.com/camunda/web-modeler/issues/13599.

### Expected Behavior

* [ ] Trim (and any input updates) are consistently applied on blur
* [ ] After blur no other updates are fired
* [ ] Text field, plain and FEEL work consistently, propagate updates down to consuming components (so business objects can be updated)

### Environment

- Host (Browser/Node version), if applicable: Chrome 130
- OS: Linux
- Library version: v3.26.3 (introduced in this version, via https://github.com/bpmn-io/properties-panel/pull/402).

abdul99ahad:
Closes #404. #309, 
Related to: https://github.com/camunda/camunda-modeler/issues/2385, https://github.com/camunda/camunda-modeler/issues/4818

## Summary
- Handling trimming input elements logic across all components (FEEL disabled (Textfield, TextArea), Text Area, Textfields)
- ~~Added test cases for trimming with debounce and without debounced~~ (Nico and I concluded that without debounce doesn't make any sense since it will never serve real world user experience) 
- Added/Adjusted test cases for trimming, debounced input, noop update on blur for all components 
- Unified input handling logic among components 
- Docs added for un documented logic (`debouncedInput`, `handleInputCallback`)
- Adjusted test cases gracefully. earlier no debounced was used violating real world user experience depiction in test cases. 
- Added integration tests in `bpmn-js-properties-panel` ([ref.](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1118))

### Proposed Changes

 
Input will be trimmed if extra spaces exists. 
Mar-26-2025 15-00-50

### Integration Test
This integration test https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1118 supports current PR 
- This ensures that when switching elements, the property will not propagate to other fields (issue found in https://github.com/camunda/web-modeler/issues/13599)
- This ensures that XML is updated when trimmed input is handled

Error
Mar-17-2025 14-22-50

Fixed 
Mar-17-2025 14-19-01

Trimming will update the input quickly and will not interfere with other elements fields

You can try this out via 
```
npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel -l bpmn-io/properties-panel#ideation-abdul
```
Follow up for : https://github.com/bpmn-io/properties-panel/pull/402, https://github.com/bpmn-io/properties-panel/pull/401

### Checklist

### DIFF
--- a/src/components/entries/FEEL/Feel.js
+++ b/src/components/entries/FEEL/Feel.js
@@ -38,6 +38,10 @@ import Tooltip from '../Tooltip';
 const noop = () => {};
 
 /**
+ * @typedef {'required'|'optional'|'static'} FeelType
+ */
+
+/**
  * @param {Object} props
  * @param {Boolean} props.debounce
  * @param {String} props.id
@@ -47,13 +51,13 @@ const noop = () => {};
  * @param {Function} props.onInput
  * @param {Function} props.onBlur
  * @param {Function} props.onError
- * @param {Boolean} props.feel
+ * @param {FeelType} [props.feel]
  * @param {String} props.value
- * @param {Boolean} props.singleLine
+ * @param {Boolean} [props.singleLine]
  * @param {Function} props.tooltipContainer
  * @param {Function | import('preact').Component} props.OptionalComponent
  * @param {Boolean} props.disabled
- * @param {Function} props.variables
+ * @param {Array} props.variables
  * @param {string} [props.placeholder]
  * @param {string | import('preact').Component} props.tooltip
  */
@@ -545,7 +549,7 @@ const OptionalFeelCheckbox = forwardRef((props, ref) => {
  * @param {String} props.description
  * @param {Boolean} props.debounce
  * @param {Boolean} props.disabled
- * @param {Boolean} props.feel
+ * @param {FeelType} [props.feel]
  * @param {String} props.label
  * @param {Function} props.getValue
  * @param {Function} props.setValue
@@ -553,7 +557,7 @@ const OptionalFeelCheckbox = forwardRef((props, ref) => {
  * @param {Function} props.validate
  * @param {Function} props.show
  * @param {Function} props.example
- * @param {Function} props.variables
+ * @param {Array} props.variables
  * @param {Function} props.onFocus
  * @param {Function} props.onBlur
  * @param {string} [props.placeholder]
@@ -665,7 +669,7 @@ export default function FeelEntry(props) {
  * @param {String} props.max
  * @param {String} props.min
  * @param {String} props.step
- * @param {Boolean} props.feel
+ * @param {FeelType} [props.feel]
  * @param {String} props.label
  * @param {Function} props.getValue
  * @param {Function} props.setValue
@@ -673,7 +677,7 @@ export default function FeelEntry(props) {
  * @param {Function} props.validate
  * @param {Function} props.show
  * @param {Function} props.example
- * @param {Function} props.variables
+ * @param {Array} props.variables
  * @param {Function} props.onFocus
  * @param {Function} props.onBlur
  */
@@ -688,7 +692,7 @@ export function FeelNumberEntry(props) {
  * @param {String} props.description
  * @param {Boolean} props.debounce
  * @param {Boolean} props.disabled
- * @param {Boolean} props.feel
+ * @param {FeelType} [props.feel]
  * @param {String} props.label
  * @param {Function} props.getValue
  * @param {Function} props.setValue
@@ -696,7 +700,7 @@ export function FeelNumberEntry(props) {
  * @param {Function} props.validate
  * @param {Function} props.show
  * @param {Function} props.example
- * @param {Function} props.variables
+ * @param {Array} props.variables
  * @param {Function} props.onFocus
  * @param {Function} props.onBlur
  * @param {string} [props.placeholder]
@@ -712,7 +716,7 @@ export function FeelTextAreaEntry(props) {
  * @param {String} props.description
  * @param {Boolean} props.debounce
  * @param {Boolean} props.disabled
- * @param {Boolean} props.feel
+ * @param {FeelType} [props.feel]
  * @param {String} props.label
  * @param {Function} props.getValue
  * @param {Function} props.setValue
@@ -720,7 +724,7 @@ export function FeelTextAreaEntry(props) {
  * @param {Function} props.validate
  * @param {Function} props.show
  * @param {Function} props.example
- * @param {Function} props.variables
+ * @param {Array} props.variables
  * @param {Function} props.onFocus
  * @param {Function} props.onBlur
  */
@@ -735,7 +739,7 @@ export function FeelToggleSwitchEntry(props) {
  * @param {String} props.description
  * @param {Boolean} props.debounce
  * @param {Boolean} props.disabled
- * @param {Boolean} props.feel
+ * @param {FeelType} [props.feel]
  * @param {String} props.label
  * @param {Function} props.getValue
  * @param {Function} props.setValue
@@ -743,7 +747,7 @@ export function FeelToggleSwitchEntry(props) {
  * @param {Function} props.validate
  * @param {Function} props.show
  * @param {Function} props.example
- * @param {Function} props.variables
+ * @param {Array} props.variables
  * @param {Function} props.onFocus
  * @param {Function} props.onBlur
  */
@@ -760,7 +764,7 @@ export function FeelCheckboxEntry(props) {
  * @param {Boolean} props.singleLine
  * @param {Boolean} props.debounce
  * @param {Boolean} props.disabled
- * @param {Boolean} props.feel
+ * @param {FeelType} [props.feel]
  * @param {String} props.label
  * @param {Function} props.getValue
  * @param {Function} props.setValue
@@ -768,7 +772,7 @@ export function FeelCheckboxEntry(props) {
  * @param {Function} props.validate
  * @param {Function} props.show
  * @param {Function} props.example
- * @param {Function} props.variables
+ * @param {Array} props.variables
  * @param {Function} props.onFocus
  * @param {Function} props.onBlur
  */
--- a/src/components/entries/FEEL/Feel.js
+++ b/src/components/entries/FEEL/Feel.js
@@ -45,6 +45,7 @@ function FeelTextfieldComponent(props) {
     label,
     hostLanguage,
     onInput,
+    onBlur,
     onError,
     placeholder,
     feel,
@@ -83,6 +84,9 @@ function FeelTextfieldComponent(props) {
     _setFocus(position + offset);
   };
 
+  /**
+   * @type { import('min-dash').DebouncedFunction }
+   */
   const handleInputCallback = useMemo(() => {
     return debounce((newValue) => {
       onInput(newValue);
@@ -92,12 +96,11 @@ function FeelTextfieldComponent(props) {
   const setLocalValue = newValue => {
     _setLocalValue(newValue);
 
-    if (typeof newValue === 'undefined' || newValue === '' || newValue === '=') {
-      handleInputCallback(undefined);
-    } else {
-      handleInputCallback(newValue);
-    }
+    // we don't commit empty FEEL expressions,
+    // but instead serialize them as <undefined>
+    const newModelValue = (newValue === '' || newValue === '=') ? undefined : newValue;
 
+    handleInputCallback(newModelValue);
   };
 
   const handleFeelToggle = useStaticCallback(() => {
@@ -127,6 +130,20 @@ function FeelTextfieldComponent(props) {
 
       // focus is behind `=` sign that will be removed
       setFocus(-1);
+    }
+  };
+
+  const handleOnBlur = (e) => {
+    const value = e.target.value;
+
+    // we trim the value, if it is needed
+    // and update input accordingly
+    if (value.trim() !== value) {
+      setLocalValue(value.trim());
+    }
+
+    if (onBlur) {
+      onBlur(e);
     }
   };
 
@@ -256,6 +273,7 @@ function FeelTextfieldComponent(props) {
             { ...props }
             popupOpen={ popuOpen }
             onInput={ handleLocalInput }
+            onBlur={ handleOnBlur }
             contentAttributes={ { 'id': prefixId(id), 'aria-label': label } }
             value={ localValue }
             ref={ editorRef }
@@ -554,7 +572,7 @@ export default function FeelEntry(props) {
     }
   }, [ value, validate ]);
 
-  const onInput = useStaticCallback((newValue) => {
+  const onInput = useCallback((newValue) => {
     let newValidationError = null;
 
     if (isFunction(validate)) {
@@ -567,7 +585,7 @@ export default function FeelEntry(props) {
     }
 
     setValidationError(newValidationError);
-  });
+  },[ element ]);
 
   const onError = useCallback(err => {
     setLocalError(err);
--- a/src/components/entries/FEEL/Feel.js
+++ b/src/components/entries/FEEL/Feel.js
@@ -37,6 +37,27 @@ import Tooltip from '../Tooltip';
 
 const noop = () => {};
 
+/**
+ * @param {Object} props
+ * @param {Boolean} props.debounce
+ * @param {String} props.id
+ * @param {Object} props.element
+ * @param {String} props.label
+ * @param {String} props.hostLanguage
+ * @param {Function} props.onInput
+ * @param {Function} props.onBlur
+ * @param {Function} props.onError
+ * @param {Boolean} props.feel
+ * @param {String} props.value
+ * @param {Boolean} props.singleLine
+ * @param {Function} props.tooltipContainer
+ * @param {Function | import('preact').Component} props.OptionalComponent
+ * @param {Boolean} props.disabled
+ * @param {Function} props.variables
+ * @param {string} [props.placeholder]
+ * @param {string | import('preact').Component} props.tooltip
+ */
+
 function FeelTextfieldComponent(props) {
   const {
     debounce,
--- a/src/components/entries/FEEL/Feel.js
+++ b/src/components/entries/FEEL/Feel.js
@@ -58,7 +58,7 @@ function FeelTextfieldComponent(props) {
     tooltip
   } = props;
 
-  const [ localValue, _setLocalValue ] = useState(value);
+  const [ localValue, setLocalValue ] = useState(value);
 
   const editorRef = useShowEntryEvent(id);
   const containerRef = useRef();
@@ -93,8 +93,7 @@ function FeelTextfieldComponent(props) {
     });
   }, [ onInput, debounce ]);
 
-  const setLocalValue = newValue => {
-    _setLocalValue(newValue);
+  const handleInput = newValue => {
 
     // we don't commit empty FEEL expressions,
     // but instead serialize them as <undefined>
@@ -110,8 +109,10 @@ function FeelTextfieldComponent(props) {
 
     if (!feelActive) {
       setLocalValue('=' + localValue);
+      handleInput('=' + localValue);
     } else {
       setLocalValue(feelOnlyValue);
+      handleInput(feelOnlyValue);
     }
   });
 
@@ -125,6 +126,7 @@ function FeelTextfieldComponent(props) {
     }
 
     setLocalValue(newValue);
+    handleInput(newValue);
 
     if (!feelActive && isString(newValue) && newValue.startsWith('=')) {
 
@@ -140,6 +142,7 @@ function FeelTextfieldComponent(props) {
     // and update input accordingly
     if (value.trim() !== value) {
       setLocalValue(value.trim());
+      handleInput(value.trim());
     }
 
     if (onBlur) {
--- a/src/components/entries/TextArea.js
+++ b/src/components/entries/TextArea.js
@@ -1,6 +1,7 @@
 import Description from './Description';
 
 import {
+  useCallback,
   useEffect,
   useLayoutEffect,
   useMemo,
@@ -50,16 +51,42 @@ function TextArea(props) {
 
   const visible = useElementVisible(ref.current);
 
+  /**
+   * @type { import('min-dash').DebouncedFunction }
+   */
   const handleInputCallback = useMemo(() => {
-    return debounce((target) => onInput(target.value.length ? target.value : undefined));
+    return debounce((newValue) => {
+      onInput(newValue);
+    });
   }, [ onInput, debounce ]);
 
-  const handleInput = e => {
-    handleInputCallback(e.target);
+  const handleInput = newValue => {
+    const newModelValue = newValue === '' ? undefined : newValue;
+    handleInputCallback(newModelValue);
+  };
 
+  const handleLocalInput = e => {
     autoResize && resizeToContents(e.target);
 
+    if (e.target.value === localValue) {
+      return;
+    }
+
     setLocalValue(e.target.value);
+    handleInput(e.target.value);
+  };
+
+  const handleOnBlur = e => {
+    const value = e.target.value;
+
+    if (value.trim() !== value) {
+      setLocalValue(value.trim());
+      handleInput(value.trim());
+    }
+
+    if (onBlur) {
+      onBlur(e);
+    }
   };
 
   useLayoutEffect(() => {
@@ -95,9 +122,9 @@ function TextArea(props) {
           monospace ? 'bio-properties-panel-input-monospace' : '',
           autoResize ? 'auto-resize' : '')
         }
-        onInput={ handleInput }
+        onInput={ handleLocalInput }
         onFocus={ onFocus }
-        onBlur={ onBlur }
+        onBlur={ handleOnBlur }
         placeholder={ placeholder }
         rows={ rows }
         value={ localValue }
@@ -157,17 +184,19 @@ export default function TextAreaEntry(props) {
     }
   }, [ value, validate ]);
 
-  const onInput = (newValue) => {
+  const onInput = useCallback((newValue) => {
     let newValidationError = null;
 
     if (isFunction(validate)) {
       newValidationError = validate(newValue) || null;
     }
 
-    setValue(newValue, newValidationError);
+    if (newValue !== value) {
+      setValue(newValue, newValidationError);
+    }
 
     setLocalError(newValidationError);
-  };
+  }, [ element ]);
 
 
   const error = globalError || localError;
--- a/src/components/entries/TextField.js
+++ b/src/components/entries/TextField.js
@@ -2,6 +2,7 @@ import Description from './Description';
 import Tooltip from './Tooltip';
 
 import {
+  useCallback,
   useEffect,
   useMemo,
   useState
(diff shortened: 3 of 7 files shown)


test/spec/components/Feel.spec.js
```diff
@@ -43,6 +43,11 @@ import FeelField, {
 insertCoreStyles();
 
 const noop = () => {};
+const debounceFn = (fn) => {
```
nikku, line 46: Consider: This adds a lot of duplication, shall we introduce a `debounce` mock in a shared helper?

src/components/entries/FEEL/Feel.js
```diff
@@ -93,7 +93,7 @@ function FeelTextfieldComponent(props) {
     });
   }, [ onInput, debounce ]);
 
-  const setLocalValue = newValue => {
+  const handleInputChange = newValue => {
```
nikku, line 96: If we go this route (which we could), then we'd want to [do that consistently across the code base](https://github.com/search?q=repo%3Abpmn-io%2Fproperties-panel%20setLocalValue&type=code). I propose to not do it (for now).

src/components/entries/TextField.js
```diff
@@ -35,20 +35,40 @@ function Textfield(props) {
...
+    return debounce((newValue) => {
+      onInput(newValue);
+    });
   }, [ onInput, debounce ]);
 
   const handleOnBlur = e => {
+    const value = e.target.value;
+
+    if (value.trim() !== value) {
+      setLocalValue(value.trim());
```
nikku, line 51: Why we don't follow the same (FEEL) pattern here? Let's align control flow across both.

abdul99ahad: It is because `setLocalValue` is different in FEEL and in normal text fields. In FEEL, it is used to set the actual value, whereas in normal fields, it is used to set local value only. So, after handling the local value, we need to handle the input

test/spec/components/Feel.spec.js
abdul99ahad: Done

src/components/entries/FEEL/Feel.js
abdul99ahad: Reverted. We can pick this up sometime in future. 

Another suggestion is we can handle setting of `value` using useEffect callback applied on `localValue`. Whenever, localValue changes, it will trigger the hook which will update the actual value. Just my two cents

test/TestHelper.js
```diff
@@ -74,4 +74,10 @@ export async function expectNoViolations(node, options = {}) {
   }
 
   expect(results.violations).to.be.empty;
-}
\ No newline at end of file
+}
+
+export function debounceFn(fn) {
```
nikku, line 79: Proposal:

* expose `createDebounce(options = { flush: noop, cancel: noop })` and use it consistently.

nikku (approved):

test/spec/components/Feel.spec.js
```diff
@@ -1887,6 +1894,7 @@ describe('<FeelField>', function() {
       // then
       await waitFor(() => {
         expect(input.value).to.equal('foo');
+        expect(flushSpy).to.have.been.called; // ensure that input changes are quick
```
nikku, line 1897: Let's not ensure flush, but instead make this a proper test:

* I input a value
* I blur the field
* Now the update is happening _immediately_
* No need for any mock.

test/spec/components/TextField.spec.js
```diff
@@ -101,8 +102,14 @@ describe('<TextField>', function() {
   it('should have no trailing spaces', async function() {
 
     // given
-    const result = createTextField({ container });
+    const flushSpy = sinon.spy();
```
nikku, line 105: To test:

* On blur, we trim
* With a pending input (debounce) we commit on blur

test/spec/components/Feel.spec.js
nikku: Specifically, to test:

* On blur, we trim
* With a pending input (debounce) we commit on blur

abdul99ahad: > With a pending input (debounce) we commit on blur

This is unclear, is this means that we go to blur having debounced function?

src/components/entries/TextField.js
abdul99ahad: Previous implementation, state is set using _setLocalValue and actual value via setLocal Value 

https://github.com/bpmn-io/properties-panel/blob/6d359c8aa2ac2cde0d8b635616bce5ff6eaf617c/src/components/entries/FEEL/Feel.js#L96-104

test/spec/components/Feel.spec.js
abdul99ahad: > Now the update is happening immediately

We can't check this without `watiFor()` since debounced function needs some time to execute flush method which will then execute the callback immediately. Do you have something else in mind? 👀

nikku (commented):
General feedback:

* Consider if/what we can unify regarding the implementation. 
* Ensure that tests capture real world use (debounced input + trim on blur)

test/spec/components/Feel.spec.js
nikku: @abdul99ahad https://github.com/bpmn-io/properties-panel/pull/405#discussion_r1999057846 indicates that you did not fully understand the different problems we need to fix:

> With a pending input (debounce) we commit on blur

Any input change is not immediately propagated, but debounced, i.e. after "input is complete". In the case of blur, we want to commit these changes immediately.

> Now the update is happening immediately

Why you cannot check this immediately? The test is simple:

* Don't mock debounce
* Trigger an input
* _[Internally a call to `props#onInput` is debounced, i.e. not yet triggered]_
* Blur the field
* _[internally `props#onInput is triggered now, because debounced function is flushed]_
* You verify `props#onInput` is called, immediately.

What am I missing?

nikku: Also worth testing:

* What happens when an input is mounted, but the underlying element changes? In that case, I'd expect any pending input to be debounced, too.

abdul99ahad: > Any input change is not immediately propagated, but debounced, i.e. after "input is complete". In the case of blur, we want to commit these changes immediately.

It took me sometime to find out why input update is happening immediately in the test cases different than in our application, the reason is missing `debouncedInput` service. The app is relying on `debouncedInput` whereas for test cases I was using the usual `debounce from min-dash`. 

> Why you cannot check this immediately? The test is simple:

I was relying on fireEvents which dispatches the event to DOM. Using element.blur(), but it was behaving very strangely and doesn't mimic our actual user interactions. ([ref.](https://testing-library.com/docs/guide-events/#focusblur))

>_Most projects have a few use cases for fireEvent, but the majority of the time you should probably use [@testing-library/user-event](https://testing-library.com/docs/user-event/intro)._ (ref. https://testing-library.com/docs/dom-testing-library/api-events)

> indicates that you did not fully understand the different problems we fixed, or I did not understand them

I understand our end goal is to resolve the input on `onBlur` immediately, so it doesn't propagate to other but the issue seems larger than this (maybe, maybe not). When I tried user interactions using `user-events`, it is still propagating value to next input even if we force resolve the debounced Input using `.flush()`. 

```js
      await userEvent.type(input, 'foo  ', { delay: 30 });
      await userEvent.tab();
      await userEvent.click(input2); // foo
```

abdul99ahad:
### Update: 

Me and @barmac debugged the test cases but couldn't reproduce the fix. We used existing fireEvents (not used user events cuz it was messing with the fake clock) to simulate the behavior. 

1. User enters the input 
2. User blurs the field 
3. User focuses the second field

**Current behavior:** 
1. Log shows that the value is set (in setValue) but spy suggests otherwise (no spy called)
2. We blur the field. blur is executed. value is trimmed. (no spy called)
3. We switch the field. We checked whether this second spy shouldn't be called (it's called)

**Ideal case:** 
1. Initial value is set after being debounced.
5. We blur the field, input is handled without debounced (forced)
6. We switch the field. (Nothing happens, empty value)

We concluded that we will try to simulate this case by writing integration testing in `bpmn-js-properties-panel`

abdul99ahad:
`bpmn-js-properties-panel` integration test for this can be found [here](https://github.com/bpmn-io/bpmn-js-properties-panel/commit/54139462740a9f4c1cdab676e8fc97b2ece3a7bd#diff-c8bdb30189e7923d74cf24e654b1b4ece981d96b3b2d775bbc1a3854030e9347)

barmac:
> `bpmn-js-properties-panel` integration test for this can be found [here](https://github.com/bpmn-io/bpmn-js-properties-panel/commit/f4804828bb9917e91a44edb80db28b2b19e9ac96)

So the test [is successful](https://github.com/bpmn-io/bpmn-js-properties-panel/actions/runs/13958494291/job/39075137411#step:8:1715) despite no fix applied in bpmn-js-properties-panel? That will not help us to verify if the solution is correct.
A proper test confirming the bug should fail until the fix is applied.
I noticed that the test is written in a way to be evergreen: https://github.com/bpmn-io/bpmn-js-properties-panel/commit/f4804828bb9917e91a44edb80db28b2b19e9ac96#diff-c8bdb30189e7923d74cf24e654b1b4ece981d96b3b2d775bbc1a3854030e9347

abdul99ahad:
> I noticed that the test is written in a way to be evergreen:

Updated: https://github.com/bpmn-io/bpmn-js-properties-panel/commit/5c9cfbd1f889dcd3be9441aebfb879750a45349f

> A proper test confirming the bug should fail until the fix is applied.

The test case is still passing despite the changes. I think it is due to the await during selection of `serviceTask2` (context is in the commit), if we do this slowly even in manual testing, the value is not propogating which makes this test true.

abdul99ahad:
PR summary updated.

src/components/entries/TextArea.js
```diff
@@ -50,16 +50,43 @@ function TextArea(props) {
...
+    const value = e.target.value;
+
+    if (value.trim() !== value) {
+      setLocalValue(value.trim());
+      handleInput(value.trim());
+    }
+
+    // we ensure that any in flight updates
+    // are being commited
+    if (handleInputCallback.flush) {
```
nikku, line 83: When would `flush` not be provided?

test/spec/components/Feel.spec.js
```diff
@@ -1871,6 +1873,47 @@ describe('<FeelField>', function() {
     });
 
 
+    it('should have no trailing spaces', async function() {
```
nikku, line 1876:
```suggestion
    it('should trim white space on blur', async function() {
```

src/components/entries/TextArea.js
abdul99ahad: When the injected debounce in props is false.

https://github.com/bpmn-io/properties-panel/blob/bc307842e1dbb9c278334356ad60963063b825ce/src/features/debounce-input/debounceInput.js#L8:L23

test/spec/components/Feel.spec.js
```diff
@@ -1871,6 +1873,47 @@ describe('<FeelField>', function() {
...
+      input.focus();
+      changeInput(input, 'foo  ');
+      input.blur();
+
+      // then
+      expect(setValueSpy).to.have.been.calledTwice;
+      expect(setValueSpy).to.have.been.calledWith('foo');
+    });
+
+    it('should have no trailing spaces - with debounce', async function() {
```
nikku, line 1894: What does this do? Is it testing the trim behavior, or is it testing the debounce behavior?

test/spec/components/Feel.spec.js
```diff
@@ -1871,6 +1873,47 @@ describe('<FeelField>', function() {
...
+
+      // given
+      const clock = sinon.useFakeTimers();
```
nikku, line 1897: Clock must be restored outside of the test, as part of an `afterEach` hook.

test/spec/components/Feel.spec.js
abdul99ahad: The complete test suite operate without debounce, `debounceInput: false`, so we won't be able to see any debounced behavior for input change or on blur. It's testing debounce behavior for blur.

We have added `flush` which immediately calls the callback in onBlur, this is what it's testing to immediately resolve the debounce when on blur.

test/spec/components/Feel.spec.js
abdul99ahad: Clock is just used within the test, inspired from...

https://github.com/bpmn-io/properties-panel/blob/bc307842e1dbb9c278334356ad60963063b825ce/test/spec/components/Feel.spec.js#L1978

nikku:
Please checkout https://github.com/bpmn-io/properties-panel/pull/405/commits/95022ab56f3578e127be43529f173e5f73e21c84 for how we want to use the clock safely, the existing tests you inspired from don't consitute "safe use".

test/spec/components/Feel.spec.js
nikku: Let's aim for tests against real world behavior. I see not using debounce in tests (which you highlight) as an anti-pattern. We don't have to clean that up today, but we can be better in our future tests.

Cf. https://github.com/bpmn-io/properties-panel/pull/405/commits/c3d3bf0902ab4887a629fe078d6835da54729cfc where I attempted to simplify your tests, now clearly capturing the scenarios you touched:

* Input is debounced
* Whatever debounce is in progress, it will be flushed on blur

nikku: In https://github.com/bpmn-io/properties-panel/pull/405/commits/e48d0e22004758bbd3fe641955bf4510be3b90a5 I add a test that verifies we only update _if necessary_.

---

We want to mirror exact these test cases, across all components we touch.

src/components/entries/TextArea.js
```diff
@@ -50,16 +50,43 @@ function TextArea(props) {
...
-
     setLocalValue(e.target.value);
+    handleInput(e.target.value);
+  };
+
+  const handleOnBlur = e => {
+    const value = e.target.value;
+
+    if (value.trim() !== value) {
+      setLocalValue(value.trim());
```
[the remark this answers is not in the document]
nikku: > Earlier, we concluded not to do it for now but will do later as part of housekeeping.

This is a mis-understanding. What I meant is that we should not, in a single place, do a certain change. But as we touch the same code path across different components anyway, lets ensure we leave the house cleaner than we entered it. ([Boy scout rule](https://deviq.com/principles/boy-scout-rule))

src/components/entries/TextArea.js
nikku: Let's ensure that whatever is returned from `debounceInput(timeout)` complies with the `DebouncedFunction` interface instead. This simplifies our way of handling this situation and makes the code maintainable. See how we [already declare the variable to be of type `DebouncedFunction`](https://github.com/bpmn-io/properties-panel/pull/405/files/2efa0f49b9322f8a6ae2404f105710cc293ae330#diff-041661313b5e904dfd0ab977c453c7cfcd888c0b3606bca5199a4f429366ca43R87). See also [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern).

nikku (changes requested):
Summarizing the results of my review:

* `setLocalValue` works differently across text input components, it is confusing, and I'd ask you to unify this as you touch related code paths anyway ([ref](https://github.com/bpmn-io/properties-panel/pull/405/files#r2014468927))
* `debounce` is of type `DebouncedFunction`, let's ensure that this is properly accounted for in our `debounceInput` helper, as well as in the tests ([ref](https://github.com/bpmn-io/properties-panel/pull/405/files#r2014483997)).
* Ensure that the tests are simple enough, meaningful, and present for all text fields, across all versions.
* I suggest that we fix our usage of `clock`, cf. https://github.com/bpmn-io/properties-panel/pull/405/commits/79d897173bb428e48c24ba0bbaee74f3cae59705 for blueprint.

src/components/entries/FEEL/Feel.js
```diff
@@ -85,7 +85,7 @@ function FeelTextfieldComponent(props) {
   };
 
   /**
-   * @type { import('min-dash').DebouncedFunction }
+   * @type { import('min-dash').DebouncedFunction || Function }
```
nikku, line 88: This works, however:

* We could instead ensure that `debounce` is always of type `DebouncedFunction`. Why wouldn't we?

abdul99ahad: This is because it depends on injected `debouncedInput` and in there, we are not restricting to only use debounce but can use without debounce as well. 
`debounce(false)`

Not to forget, current test cases that's not relying on debounce at all.

nikku: > This is because it depends on injected debouncedInput and in there, we are not restricting to only use debounce but can use without debounce as well.

I don't follow 100%.

It would be trivial for `debounceInput` to return a mock implementation adhering to `DebouncedFunction` when `debounceInput(false)` is called. If we only decide to go the current route, because of test cases, then I'd argue: Re-consider.

The tests should use standard facilities, and the least I'd expect is that they use `debounceInput(false)` to create the debounced (for testing) function.

abdul99ahad: We are not directly using `debounce` from min-dash but using it from `debouncedInput`. In there, it handles the case if user provides `false`, it will just return the function without debounce. Not really sure why it was implemented this way if the core reason of using debounce is to have `DebouncedFunction`.

https://github.com/bpmn-io/properties-panel/blob/bc307842e1dbb9c278334356ad60963063b825ce/src/features/debounce-input/debounceInput.js#L8-L22

Maybe @barmac can help clarify? 

Regarding test cases, I completely agree to simulate real world user experience, it should work with debounce, but do we want to do this within this scope?

nikku: > Not really sure why it was implemented this way if the core reason of using debounce is to have DebouncedFunction.

It was implemented that way, because we did not use any advanced features. But we are the full owners of this, and can change it in ever way we desire. One thing is for a fact: You are not supposed to use random custom `debounce`, `debounceInput` is what creates a `debounce` function for you.

test/spec/components/TextArea.spec.js
```diff
@@ -567,7 +672,8 @@ describe('<TextArea>', function() {
       });
 
       // then
-      await expectNoViolations(node);
+      expectNoViolations(node);
+      clock.tick(500);
```
nikku, line 676: Why do we tick here?

barmac: This change means that we probably ignore the assertion.

abdul99ahad: Because it shows timeout error 

```
Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

abdul99ahad: The error was due to usage of fake clock. I've fixed it.

src/features/debounce-input/debounceInput.js
```diff
@@ -5,7 +5,29 @@ import {
...
+ * const debouncedFn = debounceInput()(fn);
+ * debouncedFn(); // Executes after 300ms (default) if no further calls occur.
+ */
 export default function debounceInput(debounceDelay) {
+
+  /**
+   * Applies debounce to the provided function based on the specified delay.
+   *
+   * @param {Function} fn
+   * @return { import('min-dash').DebouncedFunction }
```
nikku, line 28: The implementation should make sure we _always_ return an instance of `DebouncedFunction`.

test/spec/components/TextArea.spec.js
```diff
@@ -38,10 +40,22 @@ const noop = () => {};
...
+  let clock;
+
   let container;
 
+  let debounce;
+
   beforeEach(function() {
     container = TestContainer.get(this);
+    clock = sinon.useFakeTimers();
+    debounce = debounceInput();
```
nikku, line 52: We don't use this anywhere but s a reference within the tests. Let's ensure the `debounceInput(false)` helper is used across the test suite, and the (standard) `debounceInput()` helper is used in the few places where needed.

nikku (changes requested):
We have to investigate the bugs that the current update caused, but also follow-up on my latest comments.

I can take that over, but not before next week.

test/spec/components/TextArea.spec.js
abdul99ahad: > Let's aim for tests against real world behavior. I see not using debounce in tests (which you highlight) as an anti-pattern.


Since the real implementation would always use debounce, do we really need to consider `debounceInput(false)` across the test suite?

nikku: We can decide to test using real debounce. Still, `debounceInput(false)` _will_ return a function, not a `DebouncedFunction`. And the code will then break.

abdul99ahad:
> Undone (via CTRL/CMD+Z) changes not reflected in properties panel:
You undo via the keyboard and old changes stay displayed in the properties panel.

This is fixed via 1d5326ede0e40be3e0bcfe24308af1deeb5d072f. Integration test is still missing (I time boxed its implementation but debounce seems to create some issues)

abdul99ahad:
> Change on blur seems to prevent selecting a different element

I tried resolving this but no progress (not wasting further time on this), the issue is because it first calls onBlur and then select the element. The inside functionality of onBlur forces the element to not instantly switch.

cc: @nikku

test/spec/components/TextArea.spec.js
abdul99ahad: > It was implemented that way, because we did not use any advanced features. But we are the full owners of this, and can change it in ever way we desire.

> The implementation should make sure we always return an instance of DebouncedFunction.

> Let's ensure the debounceInput(false) helper is used across the test suite, and the (standard) debounceInput() helper is used in the few places where needed.

I'm confused. From what I understood, there will never be a case in real world scenario where debounce is not being used, still we are using without debounce in test cases. 

If we were to limit implementation of DebouncedFunction to only return debounced function, then debounceInput(false) will be irrelevant.

nikku: `DebouncedFunction` is an interface, which provides the `cancel` and `flush` methods. Is this is where your confusion is coming from?

abdul99ahad: Nope, the confusion is regarding our debouncedInput function which supports with or without debounce. https://github.com/bpmn-io/properties-panel/blob/b0f30552e5b9edf97aa19b3c7271428bff9d91e8/src/features/debounce-input/debounceInput.js.

> The implementation should make sure we always return an instance of DebouncedFunction.

This wouldn't be possible unless we change how debounceInput behave and only allow cases to handle debounce (eliminating `debounceInput(false)`). If we were to do that, we have to rewrite all the testcases that's not relying on debounce, so we can't do that.

> Let's ensure the debounceInput(false) helper is used across the test suite, and the (standard) debounceInput() helper is used in the few places where needed.

If we continue using `debounceInput(false)`, that works fine. just two things 
1. The implementation of debounceInput would be like this which I added earlier
```js
@return { import('min-dash').DebouncedFunction || Function}
```
2. This won't represent/simulate real world scenario.

nikku:
> Perhaps it would be better to avoid updating component state on blur completely.

So this means: Only call external callbacks, don't update the component state itself? I'd try that.

abdul99ahad:
> Change on blur seems to prevent selecting a different element

Resolved this issue by avoiding the use of `flush`, as it triggered the debounce immediately. This prevented us from selecting the next element until the `onBlur` event of the previous element had fully completed.

### Resolution 

useStaticCallback in onInput uses latest props but the function is memoized, means doesn't recreate on every render. This will trigger the issue of updating the props of new element even when it's being called by the previous element blurred. 

Remove useStaticCallback, so every element has its copy of onInput. (ideal , not ideal? maybe) 

- #### useStaticCallback(value, element) 
    - Even when element is passed to have distinction for elements, the props call to `setValue` is called with latest props (and not for the current element)
- > Only call external callbacks, don't update the component state itself? I'd try that.
    - Even if we don't set component state and just call external callback (`onInput` from props), the issue would still persists due to flush function. 
- The hacky solution for using `setTimeout(() => {...},0)` would work on the same principle for async as it will not immediately update but wait for some 10ms (that's what stackoverflow peeps are saying), but then again we were using `flush` which forces the update to happen in the same cycle.

abdul99ahad:
I've re-evaluated the solution and current solution seems ideal to me.

## Known behavior
If we were to use onInput, the problem is when we undo, we first have untrimmed state and then previous state. ✅

Apr-28-2025 14-12-05

## WHYs
1. **WHY** are we setting the input on `onInput` rather than `onBlur`? Do we really need debounce? What purpose does it serve?
    1. If we set input using onBlur, we won’t be needing debounced. ❌
        Problem: If I type in another textbox, and while typing, if I undo it. It won’t remove the current text but remove the prior text. We don’t want that
        
Apr-28-2025 15-39-58
  
2. Why not trim at onInput? 
    1. If yes, why there exists onBlur? ❌
        
        Problem: If I’m writing a word with space “Activity 1”, it can trigger trim and don’t let me type space. “Activity1”
        
3. Why don’t we wait when there are trailing spaces? We should not update the input or XML if the value ends with spaces. ❌
    
    Problem: Similar to 1.

nikku (commented):
Some findings during code review. Could you follow-up and clean up the commit history (semantically structure)? Please also link the bpmn-js-properties-panel (or another place) that verifies trimming E2E, in a real-world example.

I'll continue with a manual integration test.

nikku:
I tested this, and it works. Only in rare cases, not even able to capture them on video I was able to reproduce timings where trimming does not happen; such cases involve continued typing + immediate switching and I'd argue these are edge cases.

Tested against `camunda-bpmn-js`:

```
npx @bpmn-io/sr camunda/camunda-bpmn-js -l bpmn-io/bpmn-js-properties-panel -l bpmn-io/properties-panel#ideation-abdul
```

Tested against `@camunda/linting`:

```
npx @bpmn-io/sr camunda/linting -l bpmn-io/bpmn-js-properties-panel -l bpmn-io/properties-panel#ideation-abdul
```

src/features/debounce-input/debounceInput.js
abdul99ahad: I think it depends on debounceDelay, if provided `false` it will return `Function` if not, it will provide `DebouncedFunction`. 

Imo, It should be 

```js
 * @return {import('min-dash').DebouncedFunction | Function}
```

nikku: (I'll provide an update on this PR).

nikku: Cf. https://github.com/bpmn-io/properties-panel/pull/405/commits/812c8f9ab9679e8e0b850bfe093e34e96fff6127. Check if you get editor assistance for uses of debounced functions now :slightly_smiling_face: 

Please remove remaining uses of `DebouncedFunction` from our interfaces; we should now be able to infer the type appropriately.

nikku:
From the [coverage diff](https://app.codecov.io/gh/bpmn-io/properties-panel/pull/405?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=bpmn-io) I can see that we don't test two introduced aspects (yet):

* if `onBlur` is provided, then it is called _on blur_
* if element value equals bound value, then no `onInput` is triggered

Can we add respective test coverage?

barmac:
Related to https://camunda.slack.com/archives/GP70M0J6M/p1747124050271219, I will rebase this PR so that it can be merged to stable.

barmac:
I tried this out with [linting](https://github.com/bpmn-io/properties-panel/pull/405#issuecomment-2838400768), and was not able to break it.

barmac (approved):

Issue #408, opened 2025-04-15
Decouple properties popup editor


Skaiir:
### What should we do?

Currently, the popup editor is strongly tethered to the properties panel. This is an extensibility issue which is blocking for the work we're interested in doing on web modeler to support easy feel variable testing.

There are two main parts of the popup editor that probably should be handled. One is the support of the popup editing button rendering inside of the properties panel, and secondly there is the actual rendering of the popup window.

My initial instinct is that we don't mess too much with the actual behavior. We won't do anything special with the actual popup button. We will however trigger the opening of the popup via event. 

#### Proposal

WIP
  

### Why should we do it?

To improve the architecture, support the web-modeler-acceleration feel use-case ([internal ref](https://camunda.slack.com/archives/C08J87HPYLQ)), and perhaps in the future move towards taking the popup entirely out of the library / refactoring it.

PR #409, opened 2025-04-30
feat: refactoring properties panel popup integration to support extension and extraction from the core panel library

Skaiir:
### Proposed Changes

I have annotated the commits best I can because describing the changes is too complicated without having the code visible, but in essence:

- The concept of popup is removed inside of the properties panel library and separated as much from the core as possible. What does remain (like the popup button), is now conceptualized as an expand button to be less opinionated about the implementation of the feature.
- Portal rendering of popups combined with an extension story was too complicated to handle, and is hence removed. New rendering is done by manually injecting preact render trees, and can be fully overwritten.
- Any control of popup behavior now goes through the eventBus, which allows for a complete decoupling of the rendering logic, and implementation of your own handlers.
- The module relating to the popup library was refactored.
- Tests were completely rewritten for the new structure.

#### For reviewers

Aside from the failing tests due to axe parallel execution, and review comments, this PR shouldn't expect any changes.

Untangling this was quite a challenge, so it was very hard for me to provide clean step by step commits. I would advise reviewing from the entire set of commits, and following the helper notes for guidance. But it's up to you.

### Dev notes
- [x] Test cases need to be outlined
- [ ] Tests
  - [x] New renderer module tests
  - [x] New manager module tests
  - [x] Refactor and cleanup popup tests
  - [x] Refactor and cleanup feelPopup tests
  - [x] Refactor and cleanup feel entry tests relating to popups
  - [ ] Fix axe parallel execution problems
  - [ ] Extensibility scenario test case
  - [ ] Integration-ish test case with the default popup behavior end-to-end
- [x] Issues
  - [x] The popup opens in feelers for feel in bpmn ??
  - [x] The popup now opens quite close to the element.
     - [x] Review behavior with design -> Popup will henceforth be opened in the center 
     - [x] Adjust accordingly 
  - [x] The new root element the popup is rendered in is affecting the Z-index in rendering
    - [x] Needs testing if still an issue
  - [x] Focus return on closing doesn't work
  - [x] Escape does not close the popup
  - [x] Need to finish cleaning up some popup specific terminology in the core
- [ ] Library follow ups
  - [ ] Leftover terminology migration from popup -> expand (didn't touch these due to being breaking)
     - [ ] html classes with "popup" (like the open popup button)
     - [ ] feel / feelers `popupopen` prop
  - [ ] Define expand button extensibility story, where do we take this expansion button exactly
  - [ ] Consistent entry unmount event across all fields ?
  - [ ] Move the popup functionality entirely out of the library

### Checklist

* [ ] **Brief textual description** of the changes present
* [ ] **Visual demo** attached
* [ ] Related issue linked via `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`


src/PropertiesPanel.js
```diff
@@ -217,6 +217,18 @@ export default function PropertiesPanel(props) {
     errors
   };
 
+  const [ activePopupEntryIds, setActivePopupEntryIds ] = useState([]);
```
Skaiir, line 220: :spiral_notepad:  We still need to manage the visual statue of fields that are currently popped up in the properties panel. To achieve this, we're using this piece of state and interacting with it via event, similarly to how errors currently work.

src/components/FeelPopupButton.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A reusable button component for opening popups.
+ *
+ * @param {Object} props
+ * @param {Function} props.onClick - Callback to trigger when the button is clicked.
+ * @param {string} [props.title] - Tooltip text for the button.
+ * @param {boolean} [props.disabled] - Whether the button is disabled.
+ * @param {string} [props.className] - Additional class names for the button.
+ */
+export function FeelPopupButton({
```
Skaiir, line 13: :spiral_notepad: I moved the popup button into its own component, minor thing for consistency.

src/components/entries/FEEL/Feel.js
```diff
@@ -37,7 +35,7 @@ import Tooltip from '../Tooltip';
 
 const noop = () => {};
 
-function FeelTextfieldComponent(props) {
+function FeelTextfield(props) {
```
Skaiir, line 38: :spiral_notepad:  The old context is removed, interactions with the popups are now done via the eventbus. I also cleaned up formatting a tiny bit, and removed some popup specific helpers from this class. Also removed the decorator pattern, as it was too extra.

src/features/feel-popup/FeelPopupManager.js
```diff
@@ -0,0 +1,56 @@
+export class FeelPopupManager {
```
Skaiir, line 1: :spiral_notepad: This module is injected in the parent injector and can be overwritten. This particular one manages popup lifecycle heedlessly, and fires events for the renderer to hook into.

src/features/feel-popup/FeelPopupRenderer.js
```diff
@@ -0,0 +1,70 @@
+import { render } from 'preact';
```
Skaiir, line 1: :spiral_notepad: This module is injected in the parent injector and can be overwritten. This particular one manages popup rendering, and is what will be overwritten in web modeler to implement our custom renderer.

src/features/feel-popup/components/FeelPopup.js
```diff
@@ -0,0 +1,161 @@
+import { useEffect, useRef } from 'preact/hooks';
```
Skaiir, line 1: :spiral_notepad: Mostly just moving files around into its own component folder.

src/features/feel-popup/components/FeelPopup.js
```diff
@@ -0,0 +1,161 @@
...
+
+      // close popup only if auto completion is not open
+      if (!isAutoCompletionOpen.current) {
+        onClose();
+        isAutoCompletionOpen.current = false;
+      }
+    }
+  };
+
+  useEffect(() => {
```
Skaiir, line 54: :warning: These events need reconsidering.

test/spec/components/Feel.spec.js
```diff
@@ -25,12 +25,9 @@ import {
   DescriptionContext,
```
Skaiir, line 25: :spiral_notepad: Don't look into the tests, the refactor completely breaks them.

barmac:
@philippfromme @nikku Could you have a look at this?

philippfromme:
General question: If invest that much into refactoring the popup why don't we move it out of the properties panel entirely and then verify it integrates nicely through events? This is what we did with the element templates feature as well. We just have button that triggers an event, that's it.

ev-codes:
Related to https://github.com/bpmn-io/properties-panel/issues/408

Skaiir:
@philippfromme I generally agree, and you can see I basically structured it in such a way that it would be really easy to do so. However I am going against this for now for a few reasons:

1. We're a bit time restricted for now, single library change is the quickest path to delivery
2. This needs integration in the form, bpmn, and dmn properties panels, which means it would need to become a library of its own, which again goes back to time
3. If we're fully re-implementing as a component, I think we can take the opportunity to fully get rid of the interaction issues that currently exist, but again it's an increase in scope.
4. Backwards compatibility concerns, it'd be a breaking change, maybe one that we should make, but not one that I think we have time to discuss too much yet

So I envision it as the next step, it's probably not going to be a major thing but it'll need a lot of dependency updates and I wouldn't want to fit this in our 2 week timeline.

huygur:
Hey, I am testing this implementation in Desktop Modeler and will put my findings here (I'll edit this comment if anything new comes up)

1- Focus is not preserved on the job type field after popup is closed


https://github.com/user-attachments/assets/e1775989-6625-4ad7-b365-4c1ff8f8642b

src/components/entries/FEEL/Feel.js
```diff
@@ -141,21 +143,26 @@ function FeelTextfieldComponent(props) {
...
-      id,
-      hostLanguage,
+  const handleExpandProperty = (type = 'feel') => {
+    const context = {
+      type,
+      value: feelOnlyValue,
+      variables,
       onInput: handleLocalInput,
-      position: calculatePopupPosition(containerRef.current),
+      hostLanguage,
```
Skaiir, line 152: :spiral_notepad:  Removed the popup-specific opinionated calculation functions

src/components/entries/FEEL/Feel.js
```diff
@@ -141,21 +143,26 @@ function FeelTextfieldComponent(props) {
     }
   });
 
-  const handlePopupOpen = (type = 'feel') => {
-    const popupOptions = {
-      id,
-      hostLanguage,
+  const handleExpandProperty = (type = 'feel') => {
```
Skaiir, line 146: :spiral_notepad: Generalized the terminology here, mostly to align with @nikku's future vision of how we could have similar functionality without breaking established patterns, by actually expanding the property into docked panels, instead of popups.

Expansion is a generic concept that can live in properties panel I believe, popup should not. Hence the direction. Let me know if I'm overreaching here.

src/components/entries/FEEL/Feel.js
```diff
@@ -179,6 +186,14 @@ function FeelTextfieldComponent(props) {
     setLocalValue(value);
   }, [ value ]);
 
+  // close popup on unmount
```
Skaiir, line 189: :spiral_notepad: Generic unmount events are fired here to replace the popup closing through the context we used to do.

src/components/entries/FEEL/Feel.js
```diff
@@ -752,44 +764,3 @@ export function isEdited(node) {
 function prefixId(id) {
   return `bio-properties-panel-${id}`;
 }
-
-function calculatePopupPosition(element) {
```
Skaiir, line 756: :spiral_notepad:  Move all this popup specific stuff into the helper

src/components/entries/FEEL/Feel.js
barmac: So the expansion accommodates use cases which don't fit "popup", e.g. showing a separate full screen view. I think this makes sense.

barmac:
Thanks Valentin for sharing this early, and the helpful comments. It looks that the goal of decoupling the FEEL popup is closer than ever.

> General question: If invest that much into refactoring the popup why don't we move it out of the properties panel entirely and then verify it integrates nicely through events? This is what we did with the element templates feature as well. We just have button that triggers an event, that's it.

The way I see this is that this pull request should make it possible to replace the popup. We can decide to move it entirely out of the project, but it's not necessary to accomplish the former goal. That will be a breaking change though.

philippfromme:
@barmac Makes sense.

philippfromme:
Regarding the AXE tests, so far I found that they are indeed not running sequentially at some point. The start going off the rails as soon as they are run on anything other then the test container node:

image

Skaiir:
I've marked this as ready for review and removed the must fixes:
- axe tests run fine in isolation, I think chasing what exactly is going on here is not needed, we can just skip the tests for now (as we know they run green), and follow up later
- z-index issues can easily be patched modeler side with some high specificity css rules if they need to be
- we're running out of time and are relying on this functionality for something we'd like to showcase during CamundaCon

Hope we can count on your support for getting reviews for this @barmac, @philippfromme. I will keep tackling the issues I've de-prio'd independently.

philippfromme:
Your PR introduces rendering a separate Preact tree rendered inside of the existing one. The AXE issue might be related to that.

philippfromme:
Only the newly added tests seem to fail. Interestingly, even when not rendering the popup the test fails:

image

I'l investigate further.

Updating the `axe-core` dependency didn't help.

philippfromme:
This is very likely an issue with a test that is run before the tests that are failing with the AXE error. If I run these tests first they pass.

philippfromme:
`Tooltips.spec.js` is causing the issues. Skipping it fixes them. Looking into that.

philippfromme:
AXE tests are fixed now. I will do the final review.

philippfromme:
Just so you know, I'm still working on improving the proposed solution on a [separate branch](https://github.com/bpmn-io/properties-panel/tree/popup-refactor-simplified) and will be done in a bit.

src/components/entries/FEEL/FeelPopup.js
```diff
@@ -1,304 +0,0 @@
...
-import TemplatingEditor from '../templating/TemplatingEditor';
-import { LaunchIcon } from '../../icons';
-
-export const FEEL_POPUP_WIDTH = 700;
-export const FEEL_POPUP_HEIGHT = 250;
-
-
-/**
- * FEEL popup component, built as a singleton. Emits lifecycle events as follows:
- *  - `feelPopup.open` - fired before the popup is mounted
```
philippfromme, line 25: We have removed these events which is technically a breaking change.

philippfromme: These events are being used in the Desktop Modeler: https://github.com/camunda/camunda-modeler/blob/develop/client/src/app/tabs/bpmn/modeler/features/properties-panel-keyboard-bindings/FeelPopupKeyboardBindings.js

src/PropertiesPanel.js
```diff
@@ -28,7 +28,7 @@ import {
   TooltipContext
 } from './context';
 
-import { FeelPopupRoot } from './components/entries/FEEL';
+import { ExpandedEntriesContext } from './context';
```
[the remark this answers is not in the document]
philippfromme: I've removed the context: https://github.com/bpmn-io/properties-panel/pull/409/commits/787530e6fd10a17f0f8d218651bc956ad1dc0afe

src/features/feel-popup/FeelPopupRenderer.js
```diff
@@ -0,0 +1,73 @@
...
+import { FeelPopup } from './components';
+import { query as domQuery } from 'min-dom';
+
+export class FeelPopupRenderer {
```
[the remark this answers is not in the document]
philippfromme: Simplified: https://github.com/bpmn-io/properties-panel/pull/409/commits/787530e6fd10a17f0f8d218651bc956ad1dc0afe

src/components/entries/FEEL/FeelPopup.js
philippfromme: I've added comments so we don't forget about this: https://github.com/bpmn-io/properties-panel/pull/409/commits/ae1df41e3d129934e236700fd1f1e3ff5e464caf

The issue is that even if the default renderer fires these events (they are actually fired by the component that is rendered) there is no guarantee that a custom render will fire these events.

philippfromme:
I've finished refactoring the pull request. I'm not 100% happy with the current solution as there are some regressions (cf. https://github.com/bpmn-io/properties-panel/pull/409#discussion_r2081843940). I'm testing this in Desktop and Web Modeler and will release this as an alpha so it can be integrated in Web Modeler.

philippfromme (approved):

PR #419, opened 2025-06-02
Prepare stable release

barmac:
### Proposed Changes

Closos #408 
 

### Checklist

* [ ] **Brief textual description** of the changes present
* [ ] **Visual demo** attached


barmac:
Currently, this causes bpmn-js-properties-panel test suite to hang and fail. This needs to be fixed yet.

barmac:
The test was failing on this assertion: https://github.com/bpmn-io/bpmn-js-properties-panel/blob/main/test/spec/BpmnPropertiesPanelRenderer.spec.js#L1067
When we run the `to.eql` check, mocha will try to compare `JSON.stringify` results. However, it does not work for DOM objects because of circular references. That makes `mocha` fail, so the test suite hangs. The solution is to use another assertion, e.g.
```js
      expect(feelPopup.parentNode === container, 'parent should be the popup container').to.be.true;
```

barmac:
Then, some more tests are failing due to `TypeError: feelPopup.open is not a function` => a method removed from the feature.

Skaiir:
@barmac I think you be able to substitute through `propertiesPanel.openPopup`. Although they aren't the same events, so it makes sense to intentionally break here.

nikku:
> I'm sorry but we want to avoid a breaking change if it's not absolutely necessary. 

This is the point. We have to now understand what is necessary, whether the [change is worth it](https://nikku.github.io/talks/2025-no-refactoring-issues/#know-what-and-why), at this point in time.

barmac:
Merged in a session with @philippfromme

barmac:
We verified this in camunda-modeler (bpmn, dmn). Small test changes are required regarding the rendering because they are too strict in bpmn, form (expect parent node and not just containment, e.g. https://github.com/bpmn-io/form-js/blob/7e3fcd5ea1a8937bc7a7c91972cec45600728105/packages/form-js-editor/test/spec/features/properties-panel/PropertiesPanel.spec.js#L3632; bpmn part already done).

PR #421, opened 2025-06-05
fix: use new value helpers when rerendering

barmac:
Related to https://github.com/camunda/camunda-modeler/issues/5071

### Proposed Changes

This makes sure that we use new value setters when a component is rerendered. This is required for the element templates conditional fields to work correctly, cf. https://github.com/camunda/camunda-modeler/issues/5071#issuecomment-2944421901

Verified with Desktop Modeler:

https://github.com/user-attachments/assets/3f5b9cd9-007b-469d-8772-c169575c8734

### Checklist

### DIFF
--- a/src/components/entries/FEEL/Feel.js
+++ b/src/components/entries/FEEL/Feel.js
@@ -622,7 +622,7 @@ export default function FeelEntry(props) {
     }
 
     setValidationError(newValidationError);
-  }, [ element ]);
+  }, [ element, getValue, setValue, validate ]);
 
   const onError = useCallback(err => {
     setLocalError(err);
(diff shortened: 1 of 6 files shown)


src/components/entries/FEEL/Feel.js
```diff
@@ -622,7 +622,7 @@ export default function FeelEntry(props) {
     }
 
     setValidationError(newValidationError);
-  }, [ element ]);
+  }, [ element, getValue, setValue, validate ]);
```
barmac, line 625: BTW this shows how all three text components could use shared logic. I don't want to bloat the changes today, so I'm leaving this as is.

PR #422, opened 2025-06-10
Flush debounced function

barmac:
### Proposed Changes

Currently, we ignore the debounced function in the text fields when onInput changes. This leads the scheduled calls to fire which breaks UX as unwanted changes are committed and pushed back to the entries.
In this PR, the scheduled updates are flushed which ensures smooth state transitions.

Related to https://github.com/camunda/camunda-modeler/issues/5051
Related to https://github.com/camunda/camunda-modeler/issues/5085

https://github.com/user-attachments/assets/c6352b9c-2038-4609-a5a9-3614df8ad2a6

### Checklist

### DIFF
--- a/test/spec/components/Feel.spec.js
+++ b/test/spec/components/Feel.spec.js
@@ -91,12 +91,8 @@ describe('<FeelEntry>', function() {
 
       // given
       const setValueSpy = sinon.spy();
-      const debouncedSpy = sinon.spy();
-      const debounce = () => debouncedSpy;
-
       const result = createFeelField({
         container,
-        debounce,
         setValue: setValueSpy
       });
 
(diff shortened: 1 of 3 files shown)


src/hooks/useDebounce.js
```diff
@@ -0,0 +1,17 @@
+import {
+  useCallback,
+  useEffect
+} from 'preact/hooks';
+
+export function useDebounce(callback, debounceFn) {
+  const debouncedCallback = useCallback(debounceFn(callback), [ callback, debounceFn ]);
+
+  // cancel previous function when new is created
```
philippfromme: This should be adjusted after we decided to go for flush instead of cancel.

philippfromme: Or removed...

barmac: I've updated the comment to reflect the intent.

philippfromme:
Flushing is the way to go. In the example below I set the debounce delay to 1000ms. We'd lose that input if we'd cancel instead of flushing.

electron_yeriE4t7TO

This seems to fix https://github.com/camunda/camunda-modeler/issues/5051 as well. It just works without any hickups.

electron_loXCD7erFL

barmac:
> Also, these two tests are extremely similar and I'm wondering if anyone will understand the distinction 3 months from now:

Good point. The idea was to verify that blurring happens even without any changes provided by the user. But it may be unnecessary.

barmac:
OK I will make the blur test very simple, and add separate test cases for debouncing.

barmac:
https://github.com/bpmn-io/properties-panel/pull/422/commits/d7b9e95b257d30f928636bab03d81ae7bd807be2 makes sure we call the flush function before re-rendering. There is no debounce logic in the blur tests anymore.

PR #426, opened 2025-06-16
Revert back to use static callback

barmac:
### Proposed Changes

This PR reverts back to pre-3.27 state where we used `useStaticCallback` to ensure that the model update helper is not created per each rendering of the component.
    
Alternative solution would be to make sure that depending libraries don't redefine `setValue` which we did not care about before 3.27.

Related to https://github.com/camunda/camunda-modeler/issues/5085

https://github.com/user-attachments/assets/891a7dd5-478b-4822-8327-23235f256bfc

### Checklist


barmac:
@Skaiir I assigned you for review as you suggested this solution.

PR #435, opened 2025-08-04
Default element templates to not have user edited sign

abdul99ahad:
Related to: https://github.com/camunda/camunda-modeler/issues/5170, https://github.com/bpmn-io/bpmn-js-element-templates/pull/170
### Proposed Changes

Element templates now provides `isDefault` for default properties and it will not be marked as user edited.

Aug-04-2025 13-01-50

### Checklist

### DIFF
--- a/src/components/Group.js
+++ b/src/components/Group.js
@@ -38,6 +38,7 @@ export default function Group(props) {
     entries = [],
     id,
     label,
+    isDefault,
     shouldOpen = false,
   } = props;
 
@@ -115,6 +116,7 @@ export default function Group(props) {
       <div class="bio-properties-panel-group-header-buttons">
         {
           <DataMarker
+            isDefault={ isDefault }
             edited={ edited }
             hasErrors={ hasErrors }
           />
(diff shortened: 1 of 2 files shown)


Buckwich (approved):
This works fine for normal default values. But how do we want to treat [generated values](https://docs.camunda.io/docs/components/modeler/desktop-modeler/element-templates/defining-templates/#generated-value), do you think it would be possible to detect them as well (outside of this PR)?

src/components/Group.js
```diff
@@ -38,6 +38,7 @@ export default function Group(props) {
     entries = [],
     id,
     label,
+    isDefault,
```
barmac, line 41: Is it not a duplication of the logic? We already provide `isEdited` function, and it should decide whether the marker is displayed. Why would we provide a second function to do the same?

barmac:
> do you think it would be possible to detect them as well (outside of this PR)?

For generated values, there is no default value so they will always be marked as edited.

src/components/Group.js
abdul99ahad: To differentiate between user edited and default values. (ref. https://github.com/bpmn-io/bpmn-js-element-templates/pull/170#issuecomment-3156463389)

barmac: OK but _user edited_ action is equal to `entry value is not equal to the default value`. We don't do dirty check here, but only compare the existing value to the default one.

PR #444, opened 2025-09-29
chore: change open popup icon to carbon maximise

Skaiir:
Related to https://github.com/camunda/camunda-modeler/issues/5185

### Proposed Changes

Adjusted the popup icon as it was too similar to the change element one.

We are using the carbon icon called Maximise here, I've also adjusted the padding by one pixel to get perfect centering in single line fields:

image

### Checklist

### DIFF
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -1467,7 +1467,7 @@ textarea.bio-properties-panel-input {
   top: 0;
   right: 0;
   line-height: 1;
-  padding: 2px 4px;
+  padding: 3px 4px;
   margin: 3px;
   display: none;
   background: none;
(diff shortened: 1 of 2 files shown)


Buckwich:
A thought I had was about the carbon license. By just copying the svg code we are removing the original license (Apache 2) and I think that would violate point [4.a](https://github.com/carbon-design-system/carbon/blob/main/LICENSE#L94). But I think the original icon was also a carbon icon (so already an issue) and I'm not a lawyer so I might be wrong

lmbateman:
> A thought I had was about the carbon license. By just copying the svg code we are removing the original license (Apache 2) and I think that would violate point [4.a](https://github.com/carbon-design-system/carbon/blob/main/LICENSE#L94). But I think the original icon was also a carbon icon (so already an issue) and I'm not a lawyer so I might be wrong

I just downloaded an icon from the [Carbon Icons page](https://carbondesignsystem.com/elements/icons/library/) and don't see license info in the SVG code, so I'm not sure it's necessary. But we could always add it in a comment. @Buckwich, where did you see the license info?

barmac (changes requested):
Let's add [@carbon/icons](https://www.npmjs.com/package/@carbon/icons) as a dependency of the library and import from there.

Buckwich:
I just checked their monorepo: https://github.com/carbon-design-system/carbon/blob/main/LICENSE. The figma design kits are licensed under [CC 4.0 Deed](https://creativecommons.org/licenses/by/4.0/), so also require attribution. not sure how to treat the direct downloads, from my permissive understanding they are unlicensed and therefore unusable.

PR #485, opened 2026-04-14
feat: add `JsonEditor` component

AlekseyManetov:
Related to https://github.com/camunda/camunda-modeler/issues/5855

### Proposed Changes

Added a new `JsonEditor` properties panel component with built-in JSON validation. This will be used for example data input and integrated into the [example-data-properties-provider](https://github.com/camunda/example-data-properties-provider), replacing the simple textarea currently in use.

This component can also be reused for any future properties panel inputs that require JSON formatting.

image

### Checklist

### DIFF
--- a/src/components/entries/JsonEditor.js
+++ b/src/components/entries/JsonEditor.js
@@ -1,0 +1,253 @@
+import Description from './Description';
+
+import {
+  useEffect,
+  useRef,
+  useState
+} from 'preact/hooks';
+
+import classnames from 'classnames';
+
+import { isObject } from 'min-dash';
+
+import {
+  useDebounce,
+  useError,
+  useShowEntryEvent,
+  useStaticCallback
+} from '../../hooks';
+
+import Tooltip from './Tooltip';
+
+import { EditorView, drawSelection, highlightSpecialChars, keymap, placeholder as cmPlaceholder } from '@codemirror/view';
+import { Annotation, EditorState } from '@codemirror/state';
+import { bracketMatching, defaultHighlightStyle, foldGutter, foldKeymap, syntaxHighlighting } from '@codemirror/language';
+import { closeBrackets, closeBracketsKeymap } from '@codemirror/autocomplete';
+import { defaultKeymap, history, historyKeymap } from '@codemirror/commands';
+import { json, jsonParseLinter } from '@codemirror/lang-json';
+import { linter } from '@codemirror/lint';
+
+{ /* Required to break up imports, see https://github.com/babel/babel/issues/15156 */ }
+
+const ExternalChange = Annotation.define();
+
+/**
+ * A CodeMirror based JSON editor for the properties panel.
+ *
+ * @param {object} props
+ * @param {string} props.id
+ * @param {string} props.label
+ * @param {string} [props.value]
+ * @param {Function} props.onInput
+ * @param {boolean} [props.disabled]
+ * @param {string} [props.placeholder]
+ * @param {string} [props.tooltip]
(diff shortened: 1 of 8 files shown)


nikku:
Consistency question: FEEL expression editor does not show line numbers currently - why do we show them for the JSON editor?

image

I'd vote to not show them. Errors should be inline, to allow users to navigate parse errors:

image

AlekseyManetov:
> why do we show them for the JSON editor?

This is intended to help the user navigate to errors based on the validation message behind the editor, which mentions a specific line number:

image

While users can currently find errors via highlights within the editor itself, I decided that line numbers are still would be useful. They also provide the additional benefit of folding/unfolding nested JSON levels, which is very helpful for large json.

nikku:
We have limited horizontal space - I'd not clutter it. We can expand the JSON to the full text editor, I'd expect to see line numbers there, and I do:

capture O3cKQd_optimized

If we want to collapse and expand - we don't need line numbers either:

capture 5umu2H_optimized

To the error message text: Goal is that I am able to find and fix the error - if it shows as a clear warning, then I'm happy:

image

---

(All of the examples are codemirror, our tools or [FEEL playground](https://nikku.github.io/feel-playground/?e=~3YCAkIC6gICAgICAgIC0HYhO%2FdnMhu%2BSgb1LkcD6wqlAT%2BVV4FqUn5h8gF9R0ZmDdg%2FN42CxXqqmvBS0c31aYtk7HVFwoiS%2FBBDjf37aV8A%3D&c=~3YCAkIADgYCAgICAgIC9AgCXnKyQZpl6gbPiCq6%2BkiwK21INVJZRilygv%2Bu0uuiyiSswKl%2Bz45%2F0gsastVjp%2FqG0XbVNlC3cidrUm81SwDUX51sdOkcN8sW2SMMMhZ1sOzBBA1VE1ZXXKrIv8Akkv%2Ftz0LDCFc%2F3G%2B%2FYYhBTlNK37pF492GILvshk7%2BCoS0efJt3ez%2F3sUy2OgFOPH94Tt0G&t=expression&st=true)).

AlekseyManetov:
I've update the code to not show line numbers and also simplify the error message under the input to be more generic - "JSON contains errors", this avoid overloading the interface and confusing the user with line number references

image

src/components/entries/JsonEditor.js
```diff
@@ -0,0 +1,250 @@
...
+import {
+  useDebounce,
+  useError,
+  useShowEntryEvent,
+  useStaticCallback
+} from '../../hooks';
+
+import Tooltip from './Tooltip';
+
+import { minimalSetup } from 'codemirror';
```
nikku: I propose we don't depend on `codemirror` here, as it pulls in many dependencies - [FEEL Editor](https://github.com/bpmn-io/feel-editor/blob/main/src/index.js) works without it.

src/components/entries/JsonEditor.js
```diff
@@ -0,0 +1,250 @@
...
+        doc: value,
+        extensions: [
+          minimalSetup,
+          foldGutter(),
+          bracketMatching(),
+          closeBrackets(),
+          keymap.of([
+            ...closeBracketsKeymap,
+            ...foldKeymap,
+            indentWithTab
```
nikku: Does this mirror what we have with [FEEL editor](https://github.com/bpmn-io/feel-editor) - I'd like to have it, but it is a [a11y trap](https://codemirror.net/examples/tab/)

nikku (approved):
Looks good, works as expected ⭐

PR #487, opened 2026-04-22
fix: don't show inline JSON editor error for empty value

AlekseyManetov:
### Proposed Changes

Small follow-up tweak for [PR #485](https://github.com/bpmn-io/properties-panel/pull/485) to prevent inline editor errors from being shown for empty values.

Before:

image

After:

image

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to [`@bpmn-io/properties-panel`](https://github.com/bpmn-io/
 
 ___Note:__ Yet to be released changes appear here._
 
+## 3.41.1
+
+* `FIX`: prevent inline error in JSON editor with empty value ([#487](https://github.com/bpmn-io/properties-panel/pull/487))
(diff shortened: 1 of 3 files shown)


nikku (approved):
@AlekseyManetov Looks good - please validate if 🤖 assertions for tests hold. We want to ensure that tests are green with the fix, and red (failing) without 🙃

AlekseyManetov:
> @AlekseyManetov Looks good - please validate if 🤖 assertions for tests hold. We want to ensure that tests are green with the fix, and red (failing) without 🙃

@nikku Copilot is wrong on this point. This was my first try with waitFor, but then I realized that it can’t be used for this case, because `waitFor` polls until the callback passes without throwing. For negative assertions like `expect(X).to.not.exist`, it succeeds on the very first poll immediately, because the element hasn't appeared yet. 

I've run the tests without my fix and they failed, so we're safe here.

PR #488, opened 2026-04-23
fix: fix JSON editor entry `isEdited` check from properties panel group

AlekseyManetov:
### Proposed Changes

Follow-up fix for [#485](https://github.com/bpmn-io/properties-panel/pull/485) regarding the properties panel group header 'is edited' notification. Added the `bio-properties-panel-input` class, which is used to retrieve the element for the `isEdited` callback and polish the editor local styles according to it.

Also reworked the example playground to re-render the properties panel on element changes, ensuring the 'is edited' notifications update correctly.

https://github.com/user-attachments/assets/3c7c2c99-dc10-4bfe-9ea5-89669dfc4e99

### Checklist

* [ ] Contribution __meets our [definition of done](https://github.com/bpmn-io/.github/blob/main/resources/DEFINITION_OF_DONE.md)__
* [ ] Pull request __establishes context__
  * [ ] __Brief textual description__ of the changes
  * [ ] __Screenshots or short videos__ showing UI/UX changes

### DIFF
--- a/src/components/entries/JsonEditor.js
+++ b/src/components/entries/JsonEditor.js
@@ -151,7 +151,7 @@ function JsonEditor(props) {
           { label }
         </Tooltip>
       </label>
-      <div ref={ containerRef } />
+      <div class={ classnames('bio-properties-panel-input', value && 'edited') } ref={ containerRef } />
     </div>
   );
 }
@@ -237,8 +237,7 @@ export default function JsonEditorEntry(props) {
  * Check if the JSON editor entry has been edited.
  */
 export function isEdited(node) {
-  const cmContent = node && node.querySelector('.cm-content');
-  return cmContent ? cmContent.textContent.trim().length > 0 : false;
+  return node && node.classList.contains('edited');
 }
 
 
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -569,42 +569,6 @@ textarea.bio-properties-panel-input {
   resize: vertical;
 }
 
-/**
- * JSON Editor (CodeMirror)
- */
-
-.bio-properties-panel-json-editor .cm-editor {
-  border: 1px solid var(--input-border-color);
-  border-radius: 2px;
-  background-color: var(--input-background-color);
-  font-size: var(--text-size-base);
-  font-family: var(--font-family-monospace);
-}
-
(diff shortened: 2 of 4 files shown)


src/components/entries/JsonEditor.js
```diff
@@ -151,7 +151,7 @@ function JsonEditor(props) {
           { label }
         </Tooltip>
       </label>
-      <div ref={ containerRef } />
+      <div class={ classnames('bio-properties-panel-input', value && 'edited') } ref={ containerRef } />
```
nikku, line 154: Our definition of edited is "has contents", so that works.

src/components/entries/JsonEditor.js
```diff
@@ -237,8 +237,7 @@ export default function JsonEditorEntry(props) {
  * Check if the JSON editor entry has been edited.
  */
 export function isEdited(node) {
-  const cmContent = node && node.querySelector('.cm-content');
-  return cmContent ? cmContent.textContent.trim().length > 0 : false;
+  return node && node.classList.contains('edited');
```
nikku, line 240: This looks a little bit like a hack to me - why don't look towards the editor to see if it contains contents? Or look towards the `value` which should be the source of truth?

nikku (approved):
One remark [here](https://github.com/bpmn-io/properties-panel/pull/488/changes#r3137316566) - please double-check what we derive our "isEdited" information from usually (FEEL editor, standard editors) - is it really the DOM?

Other than that, looks good. Thanks for the updates on playground and styling :star:

src/components/entries/JsonEditor.js
AlekseyManetov: @nikku I use this approach because the `JsonEditor` renders a CodeMirror `<div>`, not a native `<input>`— there's no `.value` property to check. And we also can't rely on reading `cmContent.textContent` because for the empty editor it includes the placeholder value. I thought this approach would be safer and doesn't rely on complex CodeMirror DOM element manipulation, which can break with updates. 

FEEL editor uses the same pattern — [FeelEditor.js#L188](https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/FEEL/FeelEditor.js#L188), [Feel.js#L838](https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/FEEL/Feel.js#L838).

PR #499, opened 2026-05-22
De-emphasize tooltiped element underlines

nikku:
### Proposed Changes

As per our guidelines we should assume that in more complicated scenarios most of the primary sections are documented via tooltips - however not all of them, at all times. In addition, entries may have tooltips, bringing additional details across.

This PR slightly de-prioritizes tooltiped elements underlines, so folks can focus more on the actual contents - it reduces the visual noise. At the same time, we still bring the core message across: This element offers a helpful tooltip.

image

For comparison, the before version:

image

### Try out via

```
npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel#tooltip-reworks -l bpmn-io/properties-panel#tt-underline
```

### Checklist

### DIFF
--- a/src/assets/properties-panel.css
+++ b/src/assets/properties-panel.css
@@ -1182,14 +1182,16 @@ textarea.bio-properties-panel-input {
   text-decoration: underline;
   text-decoration-style: dotted;
   text-underline-offset: 2px;
-  text-decoration-color: #888;
+  text-decoration-color: var(--color-grey-225-10-55);
   font: inherit;
   display: flex;
   justify-content: center;
   width: fit-content;
 }
 
-.bio-properties-panel-tooltip-wrapper:hover {
+.bio-properties-panel-tooltip-wrapper:hover,
+.bio-properties-panel-tooltip-wrapper:focus,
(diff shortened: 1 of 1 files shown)


nikku:
Validated this works:

* Using keyboard navigation
* Now using CSS variables to allow overriding

barinali (approved):

src/assets/properties-panel.css
```diff
@@ -1182,12 +1182,19 @@ textarea.bio-properties-panel-input {
   text-decoration: underline;
   text-decoration-style: dotted;
   text-underline-offset: 2px;
+  text-decoration-color: var(--color-grey-225-10-55);
```
barmac, line 1185: A bit late to the party but direct usage of a color will make it more difficult to theme the properties panel as you need to override `var(--color-grey-225-10-55)` in that case.

nikku: Fixed via https://github.com/bpmn-io/properties-panel/pull/500 - using a previously unused :wink: semantic color name.

PR #509, opened 2026-06-12
fix: do not render toggle when entry has no children

### LINKED ISSUE #508: CollapsibleEntry renders a non-functional toggle when entry has no children
### Describe the Bug

`CollapsibleEntry` renders a clickable header and a toggle arrow even when
the entry has no nested entries (`entries.length === 0`). Clicking the arrow
or the header flips the internal `open` state but nothing visually expands
or collapses, because the entries container is empty.

This produces a UX inconsistency:

- the arrow rotates between right / down on every click without any visible
  effect;
- screen readers announce the entry as expandable while there is nothing
  to expand;
- keyboard / pointer focus lands on a button that has no useful action.

### Steps to Reproduce

1. Render a `CollapsibleEntry` (e.g. via a list group item) with `entries: []`.
2. Click the entry header (or the arrow on its right).
3. Observe that the arrow flips orientation and the entry gets the `open`
   class, but the rendered output below is empty either way.

Minimal reproduction (inside the panel):

```js
{
  id: 'empty-item',
  component: CollapsibleEntry,
  label: 'My item',
  entries: []
}
```

### Expected Behavior

When `CollapsibleEntry` has no nested entries:

- the header should not be interactive (no toggle on click);
- the toggle arrow button should not be rendered (or should be replaced
  with non-interactive visual filler);
- assistive technology should not announce a non-functional expandable
  region.

### Environment

- Host (Browser/Node version), if applicable: Chrome 131 / Node 24 LTS
- OS: Windows 11
- Library version: 3.44.1 (latest `main`, commit `b3d6d74`)

mia-petkovic:
### Proposed Changes

`CollapsibleEntry` previously rendered an always-on toggle arrow and a clickable header, even when the entry had no nested entries (`entries.length === 0`).

Clicking the arrow or the header flipped the internal `open` state but nothing visibly expanded or collapsed, because the entries container was empty. Screen readers also announced the entry as expandable while there was nothing to expand.

This PR makes `CollapsibleEntry` only render the toggle button and react to header clicks when there is at least one nested entry to show:

- `src/components/entries/Collapsible.js`
  - introduces a local `hasEntries = entries.length > 0` flag;
  - gates the toggle arrow `<button>` rendering on `hasEntries`;
  - gates the header `onClick={ toggleOpen }` on `hasEntries` (falls back to `null`).
- `test/spec/components/Collapsible.spec.js`
  - updates the two existing tests that relied on the toggle being present to pass an `entries` array;
  - adds two new specs covering the empty-entries behaviour:
    - `should NOT render toggle button when entry has no children`
    - `should NOT toggle on header click when entry has no children`.

No public API change; the `entries` prop already defaults to `[]`, so existing call sites are unaffected.

### Checklist

- [x] Contribution meets the project's [definition of done](https://github.com/bpmn-io/.github/blob/main/resources/DEFINITION_OF_DONE.md)
- [x] Pull request establishes context
  - Link to related issue(s): Closes #508
  - Brief textual description of the changes: see above
  - Screenshots / steps to reproduce: see *Notes for reviewers* below
- [x] `npm run all` passes locally (lint + 672 tests, 1 skipped — matches `main`)

### Notes for reviewers

**Reproduction in the existing dev demo:**

1. `npm install && npm run dev`
2. In `demo/`, render a `CollapsibleEntry` with `entries: []`, e.g.:
   ```js
   {
     id: 'empty-item',
     component: CollapsibleEntry,
     label: 'My item',
     entries: []
   }
   ```
3. **Before this PR:** the arrow rotates on every header/arrow click but nothing expands or collapses.
   **After this PR:** no toggle arrow is rendered and the header is not interactive.

**Design note:** when an entry has no children the arrow is simply omitted. If you would prefer a visual placeholder (e.g. a dash/dot) to keep horizontal alignment consistent across a list of mixed entries, I'm happy to follow up — but the upstream CSS has no class for that today, so I kept the change minimal.

### DIFF
--- a/test/spec/components/Collapsible.spec.js
+++ b/test/spec/components/Collapsible.spec.js
@@ -104,7 +104,15 @@ describe('<Collapsible>', function() {
   it('should toggle open', async function() {
 
     // given
-    const { container } = createCollapsible({ container: parentContainer });
+    const { container } = createCollapsible({
+      container: parentContainer,
+      entries: [
+        {
+          id: 'entry-1',
+          component: TestEntry
+        }
+      ]
+    });
 
     const entry = domQuery('.bio-properties-panel-collapsible-entry', container);
 
@@ -122,6 +130,42 @@ describe('<Collapsible>', function() {
 
     // then
     expect(domClasses(entries).has('open')).to.be.true;
+  });
+
+
+  it('should NOT render toggle button when entry has no children', function() {
+
+    // when
+    const { container } = createCollapsible({ container: parentContainer });
+
+    const toggleButton = domQuery('.bio-properties-panel-collapsible-entry-arrow', container);
+
+    // then
+    expect(toggleButton).to.not.exist;
+  });
+
+
+  it('should NOT toggle on header click when entry has no children', async function() {
+
+    // given
+    const { container } = createCollapsible({ container: parentContainer });
+
+    const entry = domQuery('.bio-properties-panel-collapsible-entry', container);
+
+    const header = domQuery('.bio-properties-panel-collapsible-entry-header', entry);
+
+    const entries = domQuery('.bio-properties-panel-collapsible-entry-entries', entry);
+
+    // assume
+    expect(domClasses(entries).has('open')).to.be.false;
+
+    // when
+    await act(() => {
+      header.click();
+    });
+
+    // then
+    expect(domClasses(entries).has('open')).to.be.false;
   });
 
 
@@ -234,7 +278,16 @@ describe('<Collapsible>', function() {
     it('should render translated toggle button title', function() {
 
       // given
-      const { container } = createCollapsible({ container: parentContainer, translate });
+      const { container } = createCollapsible({
+        container: parentContainer,
+        translate,
+        entries: [
+          {
+            id: 'entry-1',
+            component: TestEntry
+          }
+        ]
+      });
 
       const toggleButton = domQuery('.bio-properties-panel-collapsible-entry-arrow', container);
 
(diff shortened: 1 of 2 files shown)


test/spec/components/Collapsible.spec.js
```diff
@@ -125,6 +133,42 @@ describe('<Collapsible>', function() {
...
+    const { container } = createCollapsible({ container: parentContainer });
+
+    const entry = domQuery('.bio-properties-panel-collapsible-entry', container);
+
+    const header = domQuery('.bio-properties-panel-collapsible-entry-header', entry);
+
+    const entries = domQuery('.bio-properties-panel-collapsible-entry-entries', entry);
+
+    // assume
+    expect(domClasses(entries).has('open')).to.be.false;
```
nikku, line 160:
```suggestion
expect(domClasses(entries).has('open')).to.be.false;
```

test/spec/components/Collapsible.spec.js
```diff
@@ -125,6 +133,42 @@ describe('<Collapsible>', function() {
...
+    const { container } = createCollapsible({ container: parentContainer });
+
+    const entry = domQuery('.bio-properties-panel-collapsible-entry', container);
+
+    const header = domQuery('.bio-properties-panel-collapsible-entry-header', entry);
+
+    const entries = domQuery('.bio-properties-panel-collapsible-entry-entries', entry);
+
+    // assume
+expect(domClasses(entries).has('open')).to.be.false;
```
nikku:
```suggestion
    expect(domClasses(entries).has('open')).to.be.false;
```

nikku (approved):
Thanks! Nice bug fix for an edge case ⭐

PR #517, opened 2026-06-25
test: suppress fake timer native clearTimeout warning

nikku:
## Summary

- Passes `shouldClearNativeTimers: true` to all `sinon.useFakeTimers()` calls in the test suite
- Eliminates the console warning `FakeTimers: clearTimeout was invoked to clear a native timer instead of one created by this library` that appeared during test runs, caused by `@codemirror/lint` creating a native timer before sinon's fake timers take over

## Why not remove fake timers entirely?

The tests use fake timers to control when codemirror's lint debounce fires (750 ms default delay). Removing them is viable for positive-assertion tests (`waitFor` with a long timeout), but the **negative-assertion test** ("should not show syntax warnings") would pass immediately before linting fires — giving false confidence. `shouldClearNativeTimers: true` is exactly the sinon API designed for this scenario: minimal, correct, and keeps tests fast.

---

Alternative considered: `expectEventually` (slow), `lintDelay` configuration in `*Editor` - larger change but would be consistent with JSON editor.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/spec/components/Feel.spec.js
+++ b/test/spec/components/Feel.spec.js
@@ -2936,7 +2936,7 @@ describe('<FeelEntry>', function() {
       it('should show syntax error', async function() {
 
         // given
-        const clock = useFakeTimers();
+        const clock = useFakeTimers({ shouldClearNativeTimers: true });
         const result = createFeelField({ container, getValue: () => '= ...syntax error...', feel: 'required' });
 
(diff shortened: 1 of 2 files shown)


nikku:
Minimal change.

PR #523, opened 2026-07-14
Enable pop-up editor for text areas

jarekdanielak:
### Proposed Changes

Adds the pop-up ("open in editor") experience to plain **text (textarea) fields**, which previously only existed for FEEL fields.

Related to https://github.com/camunda/camunda-modeler/issues/5123

**1. `chore(refactor): move generic popup out of feel-popup feature`**

Renames/moves `feel-popup` feature related files and dir to a generic `popup`.

**2. `fix(popup): respect popup size constants`**

The FEEL popup ignored its height constant: a CSS `min-height: 400px` overrode the inline `height` (250). Removes that CSS override and makes the JS constant the source of truth (`FEEL_POPUP_HEIGHT = 400`). Also fixes vertical centering, since `getPopupPosition()` now centers using the real height. Pre-existing bug, independent of the feature.

**3. `feat(popup): add pop-up editor for text fields`**

New `TextPopup` component, the pop-out button + placeholder on the `TextArea` entry, and renderer routing that picks FEEL vs. text via the already-existing `type` (no new config flag). Kept as a separate component from `FeelPopup` so the two can diverge later.

### Try it

```sh
npx @bpmn-io/sr -l bpmn-io/properties-panel#text-popup bpmn-io/bpmn-js-properties-panel
```

text-popup

### Checklist


nikku:
@jarekdanielak As we're on it: We've discussed in the past whether to move the whole popup feature out of the properties panel - could you evaluate what that would take?

Idea: You plug-in the popup extension - you can edit things using a popup. Similar infrastructure can be used to plug-in custom popups (feel-playground).

jarekdanielak:
@nikku good idea.

For this PR, I propose a simple "register provider" mechanism. Anyone can plug in a custom popup by registering a component for a type and firing `propertiesPanel.openPopup` with that type. Cf. https://github.com/bpmn-io/properties-panel/pull/523/changes/c3ff268fbd1cde735e4d2356c983d4d7132b9b7e

Next up, we can de-FEEL the `feelPopup` module, and extract it out of the repo if we like.

nikku:
> For this PR, I propose a simple "register provider" mechanism. 

Synchronize that with how we do registration in other places, i.e. popup menu.

jarekdanielak:
The providers are now handled via the event bus, similarly to the diagram-js popup.

nikku:
Two nit picks, adjusted:

* ID clash - https://github.com/bpmn-io/properties-panel/pull/523/commits/e9dd3e89f729c14571bbcf6441e4772f99ff0c78
* Be able to override providers via API - https://github.com/bpmn-io/properties-panel/pull/523/commits/f8048c85bac892992b2c56d86730866eda30dfc5

nikku:
Added https://github.com/bpmn-io/properties-panel/issues/528 as a follow-up.

nikku (approved):

PR #524, opened 2026-07-14
fix: support snippet placeholder navigation in FEEL popup

AlekseyManetov:
### Proposed Changes

Related to https://github.com/camunda/camunda-modeler/issues/5055

Previously in the FEEL popup, pressing `Tab` after applying a completion like `append(list, item)` moved focus out of the editor instead of jumping to the next placeholder.

The popup's focus trap intercepts `Tab` on document capture, before the editor's snippet keymap runs. `<Popup>` now accepts an `allowFocusMove(event)` predicate, and the FEEL popup returns `false` while a snippet is active (`.cm-snippetField` present), leaving the keypress to the editor. Behavior now matches the inline FEEL editor.

Also wires the FEEL popup feature in the interactive example (`npm start`), which was previously not set up.

https://github.com/user-attachments/assets/748492b8-913d-418c-8577-30f739fb14d9

### Checklist

### DIFF
--- a/src/features/feel-popup/components/FeelPopup.js
+++ b/src/features/feel-popup/components/FeelPopup.js
@@ -94,6 +94,7 @@ export function FeelPopup(props) {
       returnFocus={ false }
       closeOnEscape={ false }
       delayInitialFocus={ false }
+      allowFocusMove={ (event) => !isSnippetNavigation(event) }
       onPostDeactivate={ handleSetReturnFocus }
       height={ FEEL_POPUP_HEIGHT }
       width={ FEEL_POPUP_WIDTH }
@@ -171,4 +172,10 @@ function prefixId(id) {
 function autoCompletionOpen(element) {
   const editor = element.closest('.cm-editor');
   return editor ? editor.querySelector('.cm-tooltip-autocomplete') : null;
+}
+
+// while a snippet is active, Tab navigates its placeholders inside the editor
+function isSnippetNavigation(event) {
+  const editor = event.target.closest('.cm-editor');
+  return !!(editor && editor.querySelector('.cm-snippetField'));
 }
--- a/src/features/feel-popup/components/Popup.js
+++ b/src/features/feel-popup/components/Popup.js
@@ -28,6 +28,9 @@ const noop = () => {};
  * @param {Function} [props.onPostDeactivate]
  * @param {boolean} [props.returnFocus]
  * @param {boolean} [props.closeOnEscape]
+ * @param {(event: KeyboardEvent) => boolean} [props.allowFocusMove] -
+ *   Whether a Tab keypress may move focus. Return false to leave the keypress
+ *   to the focused element, e.g. an editor navigating snippet placeholders.
  * @param {string} props.title
  * @param {Ref} [ref]
  */
@@ -43,6 +46,7 @@ function PopupComponent(props, globalRef) {
     onPostDeactivate = noop,
     returnFocus = true,
     closeOnEscape = true,
+    allowFocusMove = () => true,
     title
   } = props;
 
@@ -103,6 +107,8 @@ function PopupComponent(props, globalRef) {
         clickOutsideDeactivates: true,
         delayInitialFocus,
         fallbackFocus: popupRef.current,
+        isKeyForward: (event) => isTab(event) && !event.shiftKey && allowFocusMove(event),
+        isKeyBackward: (event) => isTab(event) && event.shiftKey && allowFocusMove(event),
         onPostActivate,
         onPostDeactivate,
         returnFocusOnDeactivate: returnFocus,
@@ -269,6 +275,10 @@ function Footer(props) {
 }
 
 // helpers //////////////////////
+
+function isTab(event) {
+  return event.key === 'Tab';
+}
 
 function getPopupParent(node) {
   return node.closest('.bio-properties-panel-popup');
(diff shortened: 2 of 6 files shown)


src/features/feel-popup/components/Popup.js
```diff
@@ -103,6 +107,8 @@ function PopupComponent(props, globalRef) {
         clickOutsideDeactivates: true,
         delayInitialFocus,
         fallbackFocus: popupRef.current,
+        isKeyForward: (event) => isTab(event) && !event.shiftKey && allowFocusMove(event),
+        isKeyBackward: (event) => isTab(event) && event.shiftKey && allowFocusMove(event),
```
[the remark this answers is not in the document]
AlekseyManetov: True in theory, but not an issue in practice: the only consumer passes a pure DOM check, so every render's predicate behaves identically — nothing is ever stale in effect.

Also, this is not specific to this line: the whole trap config (onPostActivate, onPostDeactivate, returnFocus, delayInitialFocus) is captured the same way when the trap is created. Reworking this to track prop changes could also backfire: consumers typically pass inline functions, so a wrongly configured dependency would tear down and recreate the focus trap on every render.

This is not a real case for now, so we can address it later if required.

src/features/feel-popup/components/FeelPopup.js
```diff
@@ -172,3 +173,9 @@ function autoCompletionOpen(element) {
   const editor = element.closest('.cm-editor');
   return editor ? editor.querySelector('.cm-tooltip-autocomplete') : null;
 }
+
+// while a snippet is active, Tab navigates its placeholders inside the editor
+function isSnippetNavigation(event) {
```
nikku, line 178: General [q] - should such editor be a focus trap? I think so - independent of snippets.

Why: Because otherwise advanced editing features (indent) would otherwise not work.

If you agree, then let's revise the popup instead - it is a focus trap that can be escaped via `ESC` - https://codemirror.net/examples/tab/

nikku:
Validated (my sanity) that this works properly in the properties panel:

capture fxpF05_optimized

> The popup's focus trap intercepts Tab on document capture, before the editor's snippet keymap runs. 

Why is that and can it be "fixed" elsewhere? It sounds odd that the trap we use prevents tabbing _within_ the popup editor.

src/features/feel-popup/components/FeelPopup.js
nikku: Verified: The editor is a focus trap right now - just it does not properly handle codemirror (as per your explaination in the PR description).

https://github.com/bpmn-io/properties-panel/pull/524#issuecomment-4979661269 still applies - I'd like us to "properly" fix it, if possible.

AlekseyManetov: Currently, the `FeelEditor` doesn't include the CodeMirror `indentWithTab` extension, so the only way to use tabs is for navigating through snippet placeholders. 

If we want to enable `indentWithTab`, I guess we could extend `isSnippetNavigation` so it doesn't allow focus to move outside the editor while the editor trap is active. 

We can only enable it for the feel editor inside the popup.  Using `Esc` to exit the keyboard trap isn't an obvious action unless you've read the CodeMirror documentation. It could break standard Tab navigation through the FEEL inputs in the properties panel.

AlekseyManetov:
> Why is that and can it be "fixed" elsewhere? It sounds odd that the trap we use prevents tabbing _within_ the popup editor.

`focus-trap` listens for Tab on document in the capture phase (hardcoded), so it always runs before the editor's own keymap.

That also means it can't be fixed in the editor. focus-trap's documented extension point for this is exactly `isKeyForward/isKeyBackward` ("which keypresses count as trap navigation"), which is what we use here — the popup owns the trap, and the FEEL popup tells it that Tab during snippet navigation is not navigation. 
So this is a conflict between different focus trap behaviors - CodeMirror and the popup. The ideal solution would be to use a single focus trap for both that understands CodeMirror internals and popup elements.

nikku:
> focus-trap listens for Tab on document in the capture phase (hardcoded), so it always runs before the editor's own keymap.

Unfortunate :upside_down_face:

nikku:
From the [focus-trap](https://github.com/focus-trap/focus-trap) documentation:

> The Tab and Shift+Tab keys will cycle through the focus trap's tabbable elements but will not leave the focus trap.

So the codemirror tabble elements are not part of the popup? I don't see reason why the focus would otherwise be disallowed. Have not time to investigate right now - go forth and merge this.

I think it is not "pretty" and works around a bug that should not exist in the first place.

AlekseyManetov:
> So the codemirror tabble elements are not part of the popup?

Snippet placeholders aren't tabbable because they aren't focusable elements at all: the whole editor is a single focusable element (the contenteditable `.cm-content`), and placeholders are just styled text ranges inside it. Placeholder navigation is CodeMirror moving its selection within that one element via its own Tab keymap — browser focus never moves. So for the popup focus trap - only the entire editor is assumed as focusable element and after tab on it - focus just moved to the next element.

nikku:
Now I have everything I need :rabbit: - focus-trap intercepts tab early - codemirror cannot intercept it anymore. In both cases "tabbing to different focusable element" is not involved.

So there is likely only a few of these hacks within the system - specific behavior that we guard against.

nikku (approved):
Thanks for spending time to explain the details ⭐

PR #551, opened 2026-09-02
feat: source variables from @bpmn-io/theme

AlekseyManetov:
Related to https://github.com/bpmn-io/internal-docs/issues/1364

### Proposed Changes

The properties panel declared its own copy of the bpmn.io palette and derived every component variable from it. A theme had no semantic name to override — it had to pin `--input-border-color` and forty others by hand, and again in every other bpmn.io library.

- The private palette is replaced by the 38 semantic tokens of [`@bpmn-io/theme`](https://github.com/bpmn-io/shadcn-theme/tree/main/packages/theme) the panel reads, copied verbatim into the top of `properties-panel.css` and declared on `.bio-theme-parent`. Nothing imports the theme at runtime — the copy *is* the shipped stylesheet.
- The panel root carries `bio-theme-parent`. The FEEL popup and the tooltip render outside the panel through portals, so they carry it too.
- `npm run lint:theme`, part of `npm run all`, fails if that copy drifts from the package.
- Two hard-coded literals become variables — `--sticky-list-entry-background-color` and `--tooltip-color` — keeping their exact value.
- The button that opens the FEEL popup is fixed to paint its icon with `fill`, so its colour variables take effect at all.

### Rendered output

Every `var()` chain was resolved in both revisions and the final colours compared. **122 of 129 component variables are identical**, and none are dropped.

**Five change, all in the FEEL editor popup.** None of those colours matched anything else in the panel: three pure greys and pure black, where every other neutral in the panel is in the `225, 10%` family, and `hsl(219, 99%, 53%)`, which is form-js's blue. Each now takes the value the panel already uses for that role.

| variable | was | now | why |
| --- | --- | --- | --- |
| `--popup-background-color` | `hsl(0, 0%, 96%)` | `hsl(225, 10%, 97%)` | a pure grey used nowhere else; now the grey behind the panel's inputs and code snippets |
| `--popup-font-color` | `black` | `hsl(225, 10%, 15%)` | pure black; now the panel's base text colour |
| `--popup-title-color` | `black` | `hsl(225, 10%, 15%)` | same |
| `--feel-popup-gutters-background-color` | `hsl(0, 0%, 90%)` | `hsl(225, 10%, 90%)` | a pure grey at the lightness already used behind the FEEL indicator, so the gutter now matches it |
| `--feel-popup-close-background-color` | `hsl(219, 99%, 53%)` | `hsl(205, 100%, 40%)` | form-js's blue; now the blue the panel already uses for links, checked checkboxes and the active FEEL indicator |

**One fix.** The button that opens the FEEL popup set `color` on itself, but its icon paints with `fill` — the property every other icon in the panel uses. The icon therefore rendered pure black, and its hover state had no effect at all. Both rules now set `fill`: the icon becomes `hsl(225, 10%, 35%)`, the grey shared with the panel's other icons, and hovering it turns `hsl(205, 100%, 40%)`, as it was meant to.

**One shadow.** The dropdown button menu was two *opaque* greys, `hsl(225, 10%, 85%)` over `hsl(225, 10%, 75%)`, which paint a grey box over whatever sits behind the menu. It is now `hsla(0, 0%, 0%, 10%)` over `hsla(0, 0%, 0%, 30%)` — the same two-layer depth, translucent, so it darkens the background instead of covering it.

### Steps to try out

```bash
npx @bpmn-io/sr bpmn-io/properties-panel#shared-base-theme
```

Siblings: https://github.com/bpmn-io/diagram-js/pull/1102, https://github.com/bpmn-io/bpmn-js/pull/2497

### Checklist

  * [ ] __Screenshots or short videos__ showing UI/UX changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,11 +10,12 @@
     "preact"
   ],
   "scripts": {
-    "all": "run-s lint clean build test test:distro",
+    "all": "run-s lint lint:theme clean build test test:distro",
     "clean": "del-cli preact dist",
     "build": "rollup -c --bundleConfigAsCjs",
     "build:watch": "npm run build -- --watch",
     "lint": "eslint .",
+    "lint:theme": "bpmn-io-theme-validate src/assets/properties-panel.css",
     "start": "cross-env SINGLE_START=example npm run dev",
     "start:styles": "cross-env SINGLE_START=styles npm run dev",
     "dev": "npm test -- --auto-watch --no-single-run",
@@ -58,6 +59,7 @@
   "devDependencies": {
     "@babel/core": "^7.29.7",
     "@babel/plugin-transform-react-jsx": "^7.29.7",
+    "@bpmn-io/theme": "^0.0.0",
     "@rollup/plugin-babel": "^7.1.0",
     "@rollup/plugin-json": "^6.1.0",
     "@rollup/plugin-node-resolve": "^16.0.3",
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   "devDependencies": {
     "@babel/core": "^7.29.7",
     "@babel/plugin-transform-react-jsx": "^7.29.7",
-    "@bpmn-io/theme": "^0.0.0",
+    "@bpmn-io/theme": "^0.1.0",
     "@rollup/plugin-babel": "^7.1.0",
     "@rollup/plugin-json": "^6.1.0",
     "@rollup/plugin-node-resolve": "^16.0.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,7 @@
       "devDependencies": {
         "@babel/core": "^7.29.7",
         "@babel/plugin-transform-react-jsx": "^7.29.7",
+        "@bpmn-io/theme": "^0.1.0",
         "@rollup/plugin-babel": "^7.1.0",
         "@rollup/plugin-json": "^6.1.0",
         "@rollup/plugin-node-resolve": "^16.0.3",
@@ -578,6 +579,16 @@
       "license": "MIT",
       "dependencies": {
         "semver": "^7.7.2"
+      }
+    },
+    "node_modules/@bpmn-io/theme": {
+      "version": "0.1.0",
+      "resolved": "https://registry.npmjs.org/@bpmn-io/theme/-/theme-0.1.0.tgz",
+      "integrity": "sha512-QRvYAVRiJLTu1gyzKTBMHvcLb2wOjSpfS80c9VkEhnjX0HC+qoImd83kFUPQ0Jdzsen+b/OatY/ms7iWIHqCsw==",
+      "dev": true,
+      "license": "MIT",
+      "bin": {
+        "bpmn-io-theme-validate": "bin/validate.js"
       }
     },
     "node_modules/@camunda/feel-builtins": {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -85,6 +87,9 @@
     "mocha": "^11.8.0",
     "mocha-test-container-support": "^0.2.0",
     "npm-run-all2": "^9.0.3",
+    "postcss": "^8.5.26",
+    "postcss-cli": "^11.0.1",
+    "postcss-import": "^17.0.0",
```
barmac: Best if we don't need to postcss anything. The theme could be just copied to `dist/assets`

AlekseyManetov: Yes, that's also doable if we want to avoid adding a PostCSS dependency.

src/assets/theme.css
```diff
@@ -0,0 +1,46 @@
+/**
```
nikku: Didn't we agree we'd want to inline this in the header of the existing properties-panel.css? Most simple solution - no userland breaking change.

AlekseyManetov: I just decided to move it into a separate file to more clearly separate it from the rest of the CSS and target the validator to only the relevant part. 

Why do you think this is a breaking change? We are adding `@import './theme.css';` to `properties-panel.css`, so there shouldn't be any breaking change for users.

AlekseyManetov: I've investigated this a bit more, and yes, it could be a breaking change — at least for our Karma configs, which couldn't parse this `@import './theme.css';` with their current setup. So let's inline these styles into the package root CSS. It's not a big deal.

karma.config.js
```diff
@@ -28,9 +28,21 @@ module.exports = function(karma) {
     ],
 
     files: [
-      suite
+      suite,
+
+      // the stylesheet `@import`s the tokens as a sibling; serve them so that
+      // import resolves rather than 404s
+      {
+        pattern: 'src/assets/theme.css',
```
barmac: To remove as we wanted to inline styles instead.

package.json
barmac, line 1: This commit could be squashed; also we don't mark devdeps as `deps` commits but `chore`.

barmac: OK actually we will squash merge the entire PR.

barmac:
One caveat worth noting is that the current validation does not prevent a drift, i.e. a custom color still validates.
image

barmac:
We don't load the styles in the styles example?
image

OK the theme parent class name is missing. @copilot add it

barmac:
@copilot add FEEL fields (optional and required) to the styles test

barmac (approved):
Looks good! Let's squash-merge it :)

AlekseyManetov:
> One caveat worth noting is that the current validation does not prevent a drift, i.e. a custom color still validates.

Yes, expected — the check validates the semantic token block only: names and values in `.bio-theme-parent` must match `@bpmn-io/theme`. The rest of the stylesheet is out of scope.

Validating static colors in rest of the file would be more complex. A static value is often right (`transparent`, `currentColor`), some things we leave non-themeable on purpose (scrollbars), and a rule would have to read `color-mix()`, gradients and shorthand colour slots rather than whole declarations.

So the check covers the two mistakes that break consumers silently: using a token name the theme doesn't have, and wrong token value. Everything else is PR review now.

PR #554, opened 2026-09-04
feat: support entry diagnostics with severity

barmac:
image

image

### Proposed Changes

This PR introduced diagnostics concept to the properties panel so that one can display reports of different severity below the entries. In addition to that, the diagnostics can carry an action callback which is displayed next to the message. This will allow to remove all custom code in the properties panel shipped via https://github.com/camunda/linting-autofix.

🤖 output below:

Generalizes the per-entry **error** into a **diagnostic** carrying a severity (`info`, `warning`, `error`), where the most severe one is displayed. Severity now lives in the panel once, instead of being re-invented by each consumer.

Two things in the ecosystem motivated this:

* `linting` already computes severities and throws them away — `linting/lib/utils/properties-panel.js` drops every bpmnlint report with `category !== 'error'`, purely because the panel could not express them. Its `{ ...errors, [id]: message }` map also loses collisions silently.
* `bpmn-js-element-templates` hand-rolled the concept: it defines `{ id, severity: 'note'|'warning', message }`, maps it to panel CSS itself (`STATUS_CLASS`), composes `has-warning` / `has-error` by hand, and consumes the panel's public `useError(id)`.

Meanwhile the panel's CSS already shipped a complete `warning` tier (text colour, input border/background, focus ring, group dot, list badge) with no JS consumer at all. This wires up that half.

**API**

```js
eventBus.fire('propertiesPanel.setDiagnostics', {
  diagnostics: {
    'entry-id': [
      { severity: 'warning', message: 'Retry count exceeds recommended maximum of 10.',
        action: { label: 'Fix', tooltip: 'Set it to 10', onClick } }
    ]
  }
});
```

* `diagnostics[entryId]` is an **array**; the panel ranks it. `validate` may return a message (= `error`, unchanged) or a single diagnostic.
* Ranking is most-severe-first, ties going to the externally provided diagnostic — one shared reducer replaces eight divergent `||` chains.
* `info` renders with the existing description colour and does not mark the field.
* Group dots and list badges show the most severe diagnostic of their entries (error > warning > edited); `info` never bubbles up.
* An action button is available at every severity, styled via the new, severity-agnostic `bio-properties-panel-diagnostic-message` / `-action` classes. The previous `bio-properties-panel-error-message` / `-error-action` classes are kept alongside them (deprecated, not removed), so this is not a breaking change.

**Backwards compatibility.** `propertiesPanel.setErrors`, `ErrorsContext`, `useError` and `useErrors` are deprecated but keep working: `setErrors` is a shorthand for `setDiagnostics`, translating its string-only shape and writing into the same store — whichever event fires last, from either source, is what's shown. The legacy hooks project **error severity only** from that store, still as a string. So `useError` can never surface a warning (which would make element-templates paint `has-error` for one), and it still sees tool errors after `linting` migrates. The deprecated `errors` map stays string-only — `action` exists on diagnostics only.

**Deliberately not in this PR:** surfacing the FEEL/feelers lint warnings that are currently discarded (`Feel.js` only surfaces syntax errors, `Templating.js` filters to `severity === 'error'`). Flipping those changes every FEEL field in the modelers, so it belongs in its own PR with UX sign-off — as does removing `linting`'s `category` filter and migrating element-templates. This change is inert until a producer opts in.

**Fixes riding along**

* Checkbox entries with an error never got `has-error` — they do now.
* Templating gated on the filtered lint reports but rendered `lint[0]` from the unfiltered ones, so `[warning, error]` showed the warning's text in error styling.
* Templating was the only entry where a local lint result outranked an externally provided error; it now follows the uniform rule.

### Steps to try out

```
npm start
```

Tick **"Show diagnostics on all supported controls"** to cycle info/warning/error with actions across every entry type, and note the group dots switching between amber and red. `npm run start:styles` shows the static severity states, including "Warning with action" and "Info".

### Checklist


barmac:
I'll rebase this on top of the theme work.

barmac:
OK this is now ready for review.

barmac:
TODO: rounded buttons for the actions

src/components/entries/Checkbox.js
```diff
@@ -8,9 +8,14 @@ import {
   useState
 } from 'preact/hooks';
 
+import classnames from 'classnames';
+
 import Description from './Description';
+import DiagnosticMessage from './Diagnostic';
 import Tooltip from './Tooltip';
 
+import { ENTRY_SEVERITY_CLASS, getMostSevere } from '../util/diagnostics';
```
nikku:
```suggestion
  const diagnostic = getMostSevere(useDiagnostics(id));
```

Could instead be "useDisplayedDiagnostics(...)" - what other value does "getMostSevere" have?

src/components/entries/Checkbox.js
```diff
@@ -98,10 +103,16 @@ export default function CheckboxEntry(props) {
...
-  const error = useError(id);
+  const diagnostic = getMostSevere(useDiagnostics(id));
 
   return (
-    <div class="bio-properties-panel-entry bio-properties-panel-checkbox-entry" data-entry-id={ id }>
+    <div
+      class={ classnames(
+        'bio-properties-panel-entry',
+        'bio-properties-panel-checkbox-entry',
+        ENTRY_SEVERITY_CLASS[ diagnostic?.severity ])
```
nikku: could instead be `getDiagnosticClass(diagnostic?)` - less implementation details exposed.

src/components/entries/JsonEditor.js
```diff
@@ -215,18 +218,19 @@ export default function JsonEditorEntry(props) {
...
 
       setValue(newValue, newValidationError);
-      setLocalError(newValidationError);
+      setLocalDiagnostic(computeDiagnostic(validate, newValue, translate));
     }
   });
 
-  const error = globalError || localError;
+  // externally provided diagnostics take precedence over local ones
+  const diagnostic = getMostSevere([ ...diagnostics, localDiagnostic ]);
```
nikku, line 229: Re-used here.

src/components/entries/Tooltip.js
```diff
@@ -215,12 +223,12 @@ function Tooltip(props) {
   };
 
   return (
-    <div class="bio-properties-panel-tooltip-wrapper" tabIndex="0"
+    <div class="bio-properties-panel-tooltip-wrapper" tabIndex={ focusable ? '0' : undefined }
```
nikku, line 226: This is an entire separate feature that could be split off for easier review - also can have a dedicated changelog entry.

src/components/ListGroup.js
```diff
@@ -100,19 +107,13 @@ export default function ListGroup(props) {
...
-
-    if (!item.entries) {
-      return;
-    }
+  const allDiagnostics = useAllDiagnostics();
 
-    // also check if the error is nested, e.g. for name-value entries
-    return item.entries.some(entry => allErrors[entry.id]);
-  });
+  // also check nested diagnostics, e.g. for name-value entries
```
nikku, line 112: I wonder if such things should be split into utility functions.

src/components/Group.js
```diff
@@ -89,9 +91,9 @@ export default function Group(props) {
     return () => cancelAnimationFrame(scheduled);
   }, [ entries, setEdited ]);
 
-  // set error state depending on all entries
-  const allErrors = useErrors();
-  const hasErrors = entries.some(entry => allErrors[entry.id]);
+  // set severity state depending on all entries
+  const allDiagnostics = useAllDiagnostics();
```
nikku, line 95: Whole block could be moved to utility function - easier to read:

```
getGroupDiagnostic(...)
```

src/components/ListGroup.js
nikku:
```
getListDiagnostic(...)
```

src/PropertiesPanel.js
```diff
@@ -212,15 +215,30 @@ export default function PropertiesPanel(props) {
     getTooltipForId
   };
 
-  const [ errors, setErrors ] = useState({});
+  const [ diagnostics, setDiagnostics ] = useState({});
+
+  const onSetDiagnostics = ({ diagnostics }) => setDiagnostics(diagnostics);
```
nikku, line 220: Why don't these normalize? So we get rid of one additional in flight transformation.

CHANGELOG.md
```diff
@@ -6,7 +6,18 @@ All notable changes to [`@bpmn-io/properties-panel`](https://github.com/bpmn-io/
...
+* `FEAT`: set diagnostics through `propertiesPanel.setDiagnostics` ([#554](https://github.com/bpmn-io/properties-panel/pull/554))
+* `FEAT`: allow diagnostics to carry an action with a custom label and tooltip ([#554](https://github.com/bpmn-io/properties-panel/pull/554))
+* `FEAT`: allow `validate` to return a diagnostic ([#554](https://github.com/bpmn-io/properties-panel/pull/554))
+* `FEAT`: indicate the most severe diagnostic in group and list markers ([#554](https://github.com/bpmn-io/properties-panel/pull/554))
+* `FIX`: mark checkbox entries with errors ([#554](https://github.com/bpmn-io/properties-panel/pull/554))
+* `FIX`: report the most severe templating lint error, not the first report ([#554](https://github.com/bpmn-io/properties-panel/pull/554))
+* `FIX`: show externally provided errors over local ones in templating entries, consistent with all other entries ([#554](https://github.com/bpmn-io/properties-panel/pull/554))
+* `DEPRECATE`: `propertiesPanel.setErrors`, `ErrorsContext`, `useError` and `useErrors` in favor of their diagnostics counterparts ([#554](https://github.com/bpmn-io/properties-panel/pull/554))
```
nikku, line 17: Let's make this a dedicated section  deprecate is not a changelog marker:

```
### Deprecation

* ....
```

nikku:
[meta] PR description: I favor introductory text over walls of images - but that may just be me. 

I'd also change the description to explain *what* the feature ships, over explain *why* it ships. Why is two lines - linking to related libraries. What is the key part for future reference (i.e. from CHANGELOG).

> Fixes riding along

Yes. These could be separate commits. Or at least we should consider it for new features ("tooltip wrapper that is not focusable") - not part of the changelog, today.

nikku:
Bug in `npm start` script - errors and diagnostics are an "either or I believe" - that is how they behave - but checkbox state is not reflected accordingly:

capture yFiAeC_optimized

nikku (approved):
Overall feature design is solid - nice E2E integration. 

Look into some of the hints I added - all concerning internal use (code style) - non is public facing. 

Great work ⭐

